PR #8, opened 2014-05-28
Changing repository links and removing scientific dependency for yesod compatibility

m0dev76:
ACK. We need to look at the aeson dependency and probably bump it to make it compatible with hackage next time we release. The new version of aeson has a dependency >= 0.3 on scientific.


sha49:
ACK

jprupp:
ACK

PR #9, opened 2014-05-28
Updating aeson dependency to use scientific version >= 0.3

m0dev76:
When hackage builds haskoin, it uses aeson 0.7.0.4 and scientific 0.3 which breaks the dependencies. Aeson 0.7.0.5 and greater have correct scientific dependencies set.


jprupp:
ACK

PR #10, opened 2014-05-28
Message → Msg to avoid conflict with Protocol

jprupp:


Network/Haskoin/Stratum/Message.hs
```diff
@@ -54,7 +54,7 @@ data Id = IntId { intId :: Int }  -- ^ Id in integer form.
         deriving (Eq, Show)
 
 -- | JSON-RPC error object in v1 or v2 format.
--- Sent inside a JSONRes in case of error.
+-- Sent inside a Resoponse in case of error.
```
m0dev76: Little typo here :)
ACK otherwise

jprupp: Nice catch

PR #16, opened 2014-06-01
Use control instead of liftBaseWith and restoreM

jprupp:
Just a cosmetic impromevement, since control is the same as liftBaseWith f >>= restoreM.


m0dev76:
ACK

PR #18, opened 2014-06-04
Move script

m0dev76:
I am moving Network.Haskoin.Protocol.Script to Network.Haskoin.Script.Types and I am changing the script types of TxIn and TxOut from Script to ByteString. This is mainly because script content is ignored by block validation rules. The content is only parsed and evaluated when the scripts are being spent.


jprupp:
ACK

PR #19, opened 2014-06-04
Merge all the protocol types into one file

m0dev76:
This pull request is open to discussion. It consolidates all the types into the same file (Network.Haskoin.Protocol.Types). However, it also creates one big bloated file. Let me know what you think.


jprupp:
I say it is large, but we will be able to divide into other modules under Types if we wish to, so no big deal.

PR #21, opened 2014-06-05
Corrected multi-threaded Stratum

jprupp:
Improved multi-threaded Stratum so that it works. Moved functions from Types into Client module.


m0dev76:
ack

PR #22, opened 2014-06-06
Adding the Reject message to the protocol types

m0dev76:
The Reject type is sent to a remote peer if the local peer rejected a message for some reason. The reason and a code is sent inside that reject message. I had to move MessageCommand from Message.hs to Types.hs as the new Reject type uses it (and Message.hs imports Types.hs).


jprupp:
We'll probably have to go sweeing the data types and make them strict. I give you an ACK anyway on this change.

m0dev76:
Yeah most of them are strict. A few have probably slipped through the cracks. I'll make another pull request for those.

PR #24, opened 2014-06-08
Strict datatypes and normal-form convertion

jprupp:
Added instances of NFData for pretty much all data types. This allows for deepseq to force evaluation to normal form instead of just weak-head normal form as allowed by seq.
Most data types already had strict arguments, I set them where they were still missing.


m0dev76:
It doesn't merge automatically. You may have to rebase on master.

jprupp:
It seems good now, check the changes and ack if you're satisfied.

m0dev76:
ACK

sha49:
ACK

PR #25, opened 2014-06-10
Do not clear an input script if we failed to sign it

m0dev76:
In some cases, a script input would be cleared when signing failed. I think the correct behaviour is to leave the scrip untouched if signing it failed.


jprupp:
Oh, things were being nuked from orbit under certain conditions. Bad boy. ACK.

jprupp:
Good catch.

PR #30, opened 2014-06-20
Style changes

m0dev76:
Let me know if you agree with these style changes. I'm not sure why there are dots in there.


sha49:
For me 
   $ it is ok

jprupp:
ACK

PR #32, opened 2014-06-22
Adding separate types TxHash and BlockHash

m0dev76:
I am introducing separates types in the type system for TxHash and BlockHash. I removed Hash512, Hash256 and Hash160. the Word512, Word256 and Word160 types remain. TxHash and BlockHash are separate types so that we can make separate instances for them and deal with them differently than Word256. You can explicitly convert between them using fromIntegral. I also shuffled some of the hashing functions around where I believe they make more sense.


sha49:
ack

jprupp:
Very, very, VERY ACK!

PR #33, opened 2014-06-23
Adding a new BTC type

m0dev76:
This commit introduces a new BTC type (which was previously tracked as a Word64). The implementation is open for discussion.


jprupp:
This is a bit of a dense implementation, but it looks fine to me.

ACK

sha49:
Why you defined this constructor BTCminusOne. Is this for non-defined btc amounts?

m0dev76:
When serializing a BTC value, it needs to be within the bounds of 21M. However, there is one special case when building the hash for signing a transaction where we need a "dummy" BTC value of (-1) :: Word64 which is basically 0xfffff.... So that is way bigger than the maximum allowed value. The constructor BTCMinusOne is not exposed and should not be used except for this special case.

Another "quirck" is the fact that I allow adding two BTC values together be greater than 21M. The bound is only checked when serializing the value (with Data.Binary). This is to prevent the library from returning an error in some  very unlikely edge cases, like trying to select coins for a value of 21M bitcoins. The selection function will try to compute a target value of 21M + fees which would bust the bounds and return an error. So I think it makes sense from a type perspective to have "in-flight" computations that are > 21M BTC but if you serialize anything on the wire, then it needs to be <= 21M. 

Check the implementation in Network.Haskoin.Types.BTC and let me know if it makes sense.

m0dev76:
This solution is too complex and the bound checks are not consistent. I will implement a simpler version.

PR #34, opened 2014-06-23
Fixing Signature Decoding Problem

lornap:
decodeSig fails on a particular signature in the blockchain due to the assertions of the DER length.

The txid is fb0a1d8d34fa5537e461ac384bac761125e1bfa7fec286fa72511240fa66864d and the signature is in the input sigScript.

This pull request contains the removal of the assertions as well as a unit test illustrating and testing the problem and solution.

### DIFF
--- a/Network/Haskoin/Crypto/BigWord.hs
+++ b/Network/Haskoin/Crypto/BigWord.hs
@@ -286,8 +286,6 @@ instance Binary (BigWord ModN) where
         unless (t == 0x02) (fail $
             "Bad DER identifier byte " ++ (show t) ++ ". Expecting 0x02" )
         l <- getWord8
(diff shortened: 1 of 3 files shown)


jprupp:
Hey lornap, thanks for the catch. We are reviewing the code.

lornap:
No problem.  Let me know if anything seems incorrect.

m0dev76:
Thanks a lot for taking the time to test this. I am being too strict indeed with DER encodings when decoding signatures. I need to look further into the reference implementation to understand which encodings are valid and which are not, to prevent the reverse problem of haskoin accepting a transaction that the reference implementation rejects. 

However, I believe the correct place to do these checks is after decoding because the checks may change from block to block. This patch is a good first step in that direction.

PR #35, opened 2014-06-23
Add methods hashSha1 and hashSha1BS

OttoAllmendinger:
needed by OP_SHA1 in script evaluation

hash functions might need a rename in order to be more consistent


jprupp:
ACK

Thanks Otto! :+1:

sha49:
ack

m0dev76:
Thanks for the patch. I agree with you that we will have to rename the hash functions at some point to better reflect their function.

PR #36, opened 2014-06-24
Code to execute Bitcoin scripts

lornap:
Includes
- Code parser to a statement tree in StatementTree.hs
- Monadic execution of OPs in Exec.hs
- Addition of some missing OP_\* in Script/Types.hs
- New dependency on parsec
- Some basic tests on stack value functions

Able to verify the transaction scripts in the first ~180 000 blocks in the block chain.


OttoAllmendinger:
I think we have a little overlap...

https://github.com/OttoAllmendinger/haskoin/tree/full-script

I'm testing against the test suite here: https://github.com/OttoAllmendinger/haskoin/blob/full-script/tests/data/script_valid.json and it's currently passing all valid tests. You have to be careful about multiple alternating `OP_ELSE` in particular, as far as I can see your code does not handle this correctly

However you have a correct implementation of `OP_CHECKSIG` and `OP_CHECKMULTISIG`  where I don't

How about we combine our efforts?

jprupp:
Hey,

Thanks for the patch. It looks good. However, Otto has been working on the script interpreter for awhile. Combining the efforts that went to both branches would be great.

We have an IRC channel #haskoin on freenode. Feel free to join. We have most of our technical discussions there.

Cheers!

lornap:
Hi all, 

Thanks for taking a look at my code. I'm happy to work with you in any way, modifying my stuff to fit better, contributing to Otto's branch, or anything I'm between. 

That said, I will try to sell my approach a little bit. I think an initial parsing of if statements to a tree is the way to go since, as branching OPs, they are very different than the others. Forming trees of statements is common in compilers/interpreters and, while Bitcoin scripts are a simple language, it makes sense here too. In that vein, Otto could you explain further the issue with else statement handling? Whatever the problem, I'm confident that it can be solved with a line or two in StatementTree.hs thanks to parsec's power and flexibility with DSLs and ASTs. 

I'll dust off Irsii and join #haskoin when I can, but with my timezone and free hours, I'm guessing github or email may work best for me. 

I'm really liking the haskoin library. You guys are doing great work!

jprupp:
We'll let you two guys discuss it and decide what gets merged by consensus. After all, you two have given this part of the code more thought than we have.

In the future, it would be great to get pull requests more often into master, so that we minimize the risk of having two people working on the same feature.

OttoAllmendinger:
About the multiple `OP_ELSE`s: an `OP_IF`-branch can have multiple `ELSE`-branches, each `ELSE` inverts the condition: https://github.com/bitcoin/bitcoin/blob/0.9.2/src/test/data/script_valid.json#L43 Should also be trivial to implement in Parsec. 

I'm almost finished though, the only thing missing is testing against `script_invalid.json` and the `OP_CHECKSIG` hooks. I think it's simply more efficient to integrate your checkSig functionality into my approach than the other way around?

lornap:
Hey Otto,

Wow.  An interesting quirk of the if statement specification.

Porting over the checkCode should be do-able.  One thing required is that the full Tx, input index and output script needs to be available.  I do this with a ReaderT.  We'd need to re-jigger some of your code to expose it.  I think your SigCheck might be a wrong approach.

What are your thoughts on building an AST form the if statements?  I think it's a clean, traditional and maintainable (of course, I'm a little biased).

Moving OP code logic is pretty straight forward in either direction.  I think an important thing to decide is which architecture to go with.  Thoughts?

OttoAllmendinger:
Well building an AST is great if you have a grammar, however Bitcoin Transaction Scripts does not have a formal grammar and is rather defined by opcodes and their effect on a stack machine. It's the way the C++ implementation does it (and most other implementations I've seen), and why I chose the stack machine model. Using parsec seems to work in this instance, but it's by accident I think. 

Consider this program

`RETURN IF IF IF IF` or 
`RETURN ENDIF ENDIF ENDIF`

They both evaluate to False, but they are "syntactically valid". The spec could have an opcode `SUCCESS` that forces the script to abort and evaluate to True, and then `1 IF SUCCESS IF IF IF` would be valid but very difficult to parse and evaluate with Parsec.

So I lean towards following the C++ code (being the spec in essence), which does no parsing at all and just evaluates opcodes one-by-one.

lornap:
Hey Otto,

I think being safely compliant with the Bitcoin core spec implementation is a good argument.  I can think of solutions for AST to work but, your'e right, it might not be worth it considering that being close to the C++ implementation is quite important here.

Let's go with your branch.

To implement the signature checking OPs, the OP code logic needs the spending Tx and input/redeemScript index as well as the output script we are trying to satisfy.  It needs this to generate the signature with which to compare.  If you expose these, I can easily migrate over the *CHECKSIG code.

Happy to move this discussion off this pull-request.

OttoAllmendinger:
Great! I'm a little busy with exam preparation but I should be able to finish this over the next two weeks. Let's coordinate on the issue tracker on my fork:

https://github.com/OttoAllmendinger/haskoin/issues/1

PR #41, opened 2014-06-26
Removing signature placeholders in partially signed transaction inputs

m0dev76:
In the spirit of interoperability with other libraries, this patch removes the signature placeholder we had in transaction inputs when an input was partially signed. Some changes had to be made to code that decides if an input is partial or complete. In some cases, the information is not available if we don't have access to the previous outputs of a transction (ex: regular SpendMulSig input).


jprupp:
ACK

sha49:
ack

hope we come back this changes some day :)

PR #42, opened 2014-06-26
Removing BuildMonad

m0dev76:
Transaction signing functions now return a status flag along with the signed transaction.


sha49:
ack

i like deleting code :)

jprupp:
ACK

More than a hundred lines less. Great job!

PR #43, opened 2014-06-26
Non-standard inputs must be considered complete

m0dev76:
As we have no introspection into non-standard inputs, they should not affect negatively the status of the whole transaction. We simply ignore them by setting the status to SigComplete.


jprupp:
ACK

sha49:
ack

PR #44, opened 2014-06-26
base58.hs haskoiner reading

sha49:
Basically, now the en(de)coding to base58 is done by readInt and showIntAtBase from Numeric. Im not sure if its faster but is easier to read. Moreover I created the function decodeBase58I to be consistent with having all the encoding functions and its inverse.

### DIFF
--- a/Network/Haskoin/Crypto/Base58.hs
+++ b/Network/Haskoin/Crypto/Base58.hs
@@ -30,7 +30,6 @@ import Data.Aeson
 
 import qualified Data.ByteString as BS
 import qualified Data.ByteString.Char8 as B8
-import qualified Data.Map.Strict as M
 import qualified Data.Text as T
 
 import Network.Haskoin.Crypto.BigWord
@@ -54,7 +53,7 @@ encodeBase58I i = fromString $ showIntAtBase 58 f (fromIntegral i) ""
 decodeBase58I :: BS.ByteString -> Maybe Integer
 decodeBase58I s = case go of 
     Just (r,[]) -> Just r
-    otherwise   -> Nothing
+    otherwise -> Nothing
   where
     c = b58' . fromIntegral . ord
     p = isJust . c 
--- a/Network/Haskoin/Crypto/Base58.hs
+++ b/Network/Haskoin/Crypto/Base58.hs
@@ -29,38 +29,53 @@ import Data.Aeson
 import qualified Data.ByteString as BS
 import qualified Data.Map.Strict as M
 import qualified Data.Text as T
---import qualified Data.ByteString.Char8 as B8
 
 import Network.Haskoin.Crypto.BigWord
 import Network.Haskoin.Crypto.Hash
 import Network.Haskoin.Util 
+import Data.Char
+import Numeric
+import Data.String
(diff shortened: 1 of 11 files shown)


Network/Haskoin/Crypto/Base58.hs
```diff
@@ -32,32 +37,29 @@ import Network.Haskoin.Crypto.BigWord
...
-b58String = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
-
 b58Data :: BS.ByteString
-b58Data = BS.pack $ map (fromIntegral . ord) b58String
-
-b58Data' :: M.Map Word8 Int
-b58Data' = M.fromList $ zip (BS.unpack b58Data) [0..57]
+b58Data = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz"
 
 b58 :: Word8 -> Word8
```
jprupp, line 42: Are you sure this needs to take a Word8 as input. If it can take an Int, then you do not need fromIntegral in the function. Change it and see if anything breaks.

Network/Haskoin/Crypto/Base58.hs
```diff
@@ -32,32 +37,29 @@ import Network.Haskoin.Crypto.BigWord
...
-encodeBase58I i
-    | i >= 0 = go BS.empty i
-    | otherwise = error "encodeBase58 is not defined for negative Integers"
-  where 
-    go acc 0 = acc
-    go acc n = go (BS.cons (fromIntegral b) acc) q
-      where 
-        (q,r) = n `quotRem` 58
-        b     = b58 $ fromIntegral r
+encodeBase58I i = fromString $ showIntAtBase 58 f (fromIntegral i) ""
```
jprupp, line 49: If you make this function take Integral a instead of Integer, you do not need the fromIntegral in this line.

Network/Haskoin/Crypto/Base58.hs
```diff
@@ -32,32 +37,29 @@ import Network.Haskoin.Crypto.BigWord
...
+  where
+    f = chr . fromIntegral . b58 . fromIntegral
```
jprupp, line 51: If b58 takes an Int as I suggested, you do not need the second call to fromIntegral.

Network/Haskoin/Crypto/Base58.hs
```diff
@@ -32,32 +37,29 @@ import Network.Haskoin.Crypto.BigWord
...
+
+decodeBase58I :: BS.ByteString -> Maybe Integer
```
jprupp, line 53: See if you can also fix the signatures to reduce the calls to fromIntegral here and elsewhere.

sha49:
The problem is that b58 function is used in two different context. When its used with BS.map should be word8 -> word8 and when its used with showInAtBase should be composed with chr to be something of type int -> char ... I think it is ok right now

jprupp:
ACK

m0dev76:
ACK, but you probably need to rebase this on master before you can merge it.

PR #45, opened 2014-06-26
Removing SigNeedPrevOut status 

m0dev76:
I'm also returning SigComplete if no SigInput is provided by the user. This is the best we can do with limited information.


jprupp:
ACK

sha49:
ACK

PR #49, opened 2014-06-28
Simplified signatures

jprupp:
Superseded by pull request #50.

I believe transaction signature was too convoluted in the original code. Instead of trying to work iteratively to refactor that code, I decided to roll my own, replacing the way transaction signatures are made and validated. The main goal of this code is clarity: to be easier to read, and precision. I spent part of the week mapping out the behaviour of bitcoind's signing code in order to come up with this scheme.

Some highlights are:
- No separate code or types for P2SH transactions
- Removal of signature status type, it is enough to use a boolean value to convey transaction signature status
- Code common to monadic (SecretT) and deterministic signing code factored out
- Signing functions split according to function and layer of abstraction, with intuitive naming and documentation
- Interface of higher-level function remains mostly unchanged
- Signature validation to determine whether transaction is completely signed is performed at the highest layer
- Reduction and confinement of code that makes contextual deductions, in favor of explicit parameters

My branch in terms of lines of code is a tad longer than plaprade's similar modifications by a small number of lines, but this code has more documentation, and is written in a sparse style for readability.

### DIFF
--- a/Network/Haskoin/Transaction/Builder.hs
+++ b/Network/Haskoin/Transaction/Builder.hs
@@ -3,8 +3,6 @@ module Network.Haskoin.Transaction.Builder
 , buildTx
 , buildAddrTx
 , SigInput(..)
-, SigStatus(..)
-, isSigInvalid
 , signTx
 , signInput
 , detSignTx
@@ -179,20 +177,6 @@ data SigInput
 instance NFData SigInput where
     rnf (SigInput o p h b) = rnf o `seq` rnf p `seq` rnf h `seq` rnf b
 
-data SigStatus
-    = SigComplete
-    | SigPartial
-    | SigInvalid !String
-    deriving (Eq, Read, Show)
-
-instance NFData SigStatus where
-    rnf (SigInvalid s) = rnf s
-    rnf x = x `seq` ()
-
-isSigInvalid :: SigStatus -> Bool
-isSigInvalid (SigInvalid _) = True
-isSigInvalid _              = False
-
 -- | Prepare input for signing.
 signInputParams :: Tx                 -- ^ Transaction to sign
                 -> Int                -- ^ Input index to sign
@@ -208,7 +192,7 @@ signInputParams tx ini prv si = (pub, pks, sgs, sh, p2s)
     pks = either error id $ decodeOutput $ sigDataOut si
     sgs = decode' $ scriptInput (txIn tx !! ini)
 
--- | Replace input script in a transaction. Use after signing.
+-- | Replace input script in a transaction. Used internally by signing code.
 replaceInput :: Tx           -- ^ Transaction
              -> Int          -- ^ Input index
              -> ScriptOutput -- ^ PkScript
@@ -258,8 +242,7 @@ addSig tx ini pks scr pub ts p2s = case pks of
       ops = scriptOps scr
       scr' = if p2s && not (null ops) then Script (init ops) else scr
 
--- | Sign a transaction input with a private key. Return a new transaction with
--- the signature applied to the specified input.
+-- | Sign a transaction input with a private key. Monadic version.
 signInput :: Monad m
           => Tx              -- ^ Transaction to sign
           -> Int             -- ^ Input index to sign
@@ -274,8 +257,7 @@ signInput tx ini si prv = do
     msg = txSigHash tx (encodeOutput pks) ini sh
     (pub, pks, sgs, sh, p2s) = signInputParams tx ini prv si
 
--- | Sign a transaction input with a private key. Return a new transaction with
--- the signature applied to the specified input.
+-- | Sign a transaction input with a private key. Deterministic version.
 detSignInput :: Tx          -- ^ Transaction to sign
              -> Int         -- ^ Input index to sign
              -> SigInput    -- ^ Signature parameters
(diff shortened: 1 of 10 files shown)


Network/Haskoin/Transaction/Builder.hs
```diff
@@ -163,159 +162,172 @@ buildTx xs ys = mapM fo ys >>= \os -> return $ Tx 1 (map fi xs) os 0
...
+                -> PrvKey             -- ^ Key to sign input with
+                -> SigInput           -- ^ Signature parameters
+                -> (PubKey, ScriptOutput, Script, SigHash, Bool)
+                -- ^ (Public key, pkscript, sigscript, hashtype, P2SH)
+signInputParams tx ini prv si = (pub, pks, sgs, sh, p2s)
+  where
+    pub = derivePubKey prv
+    sh  = sigDataSH si
+    p2s = sigIsP2SH si
+    pks = either error id $ decodeOutput $ sigDataOut si
```
m0dev76: I do not believe that providing a non-standard output here is a programming error. If it is a programming error, then why let the user make the error in the first place? If you believe it is a programming error, then you should only accept ScriptOutput in the SigInput type, but the library will lack generality. The users of this library should be able to recover gracefully from this error, especially if it is also pure code. Returning False as a signing status could be appropriate in this case, or return a more complete status like SigInvalid with a message.

Network/Haskoin/Transaction/Builder.hs
```diff
@@ -163,159 +162,172 @@ buildTx xs ys = mapM fo ys >>= \os -> return $ Tx 1 (map fi xs) os 0
...
+    sgs = decode' $ scriptInput (txIn tx !! ini)
+
+-- | Replace input script in a transaction. Used internally by signing code.
+replaceInput :: Tx           -- ^ Transaction
+             -> Int          -- ^ Input index
+             -> ScriptOutput -- ^ PkScript
+             -> ScriptInput  -- ^ SigScript
+             -> Bool         -- ^ P2SH flag
+             -> Tx           -- ^ Modified transaction
```
m0dev76, line 206: I have a helper called updateIndex in Network.Haskoin.Util. Have a look if it can simplify this code.

Network/Haskoin/Transaction/Builder.hs
```diff
@@ -163,159 +162,172 @@ buildTx xs ys = mapM fo ys >>= \os -> return $ Tx 1 (map fi xs) os 0
...
+        PayPK _         -> SpendPK ts
+        PayMulSig _ r   -> case null (scriptOps scr') of
+            True  -> msAddSig tx ini pks (SpendMulSig [] r) ts
+            False -> case decodeInput pks scr' of
+                Right (sgs, _, _) -> msAddSig tx ini pks sgs ts
+                Left e -> error $ "addSig: " ++ e
+        PayScriptHash _ -> undefined
+    where
+      ops = scriptOps scr
+      scr' = if p2s && not (null ops) then Script (init ops) else scr
```
m0dev76: This is a little bit of a hack no? You are chopping off part of the script and re-interpreting it as something else. The script as a whole is a P2SH input and should be decoded as such I believe. Otherwise, you are either duplicating functionality, or introducing errors. In this case, you do not check anything about the bit you chop off. What if it's not a OP_PUSHDATA? You don't know if the script is even a valid P2SH as a whole.

Network/Haskoin/Transaction/Builder.hs
```diff
@@ -163,159 +162,172 @@ buildTx xs ys = mapM fo ys >>= \os -> return $ Tx 1 (map fi xs) os 0
...
+       -> TxSignature   -- ^ Transaction signature
+       -> Bool          -- ^ P2SH
+       -> ScriptInput   -- ^ SigScript
+addSig tx ini pks scr pub ts p2s = case pks of
+        PayPKHash _     -> SpendPKHash ts pub
+        PayPK _         -> SpendPK ts
+        PayMulSig _ r   -> case null (scriptOps scr') of
+            True  -> msAddSig tx ini pks (SpendMulSig [] r) ts
+            False -> case decodeInput pks scr' of
+                Right (sgs, _, _) -> msAddSig tx ini pks sgs ts
```
m0dev76: Due to the pattern matching in msAddSig, if sgs is not of type SpendMulSig this will return an error. I do not believe that this is a programming error and it should be allowed to recover gracefully through a meaningful status code (False or SigInvalid). I do not expect users to do full structural validation of transactions coming from elsewhere before calling this library.

Network/Haskoin/Transaction/Builder.hs
```diff
@@ -163,159 +162,172 @@ buildTx xs ys = mapM fo ys >>= \os -> return $ Tx 1 (map fi xs) os 0
...
+            -- ^ Returns tuple with parameters required to sign inputs
+getSigParms (Tx _ ti _ _) sigis keys =
+    [ (si, input, prvM)
+    | (si, input) <- findSigInput sigis ti
+    , let out = sigDataOut si
+    , let sks = map Just . either e id $ sigKeys out keys
+    , prvM <- if null sks then [Nothing] else sks
+    ]
+  where
+    e s = error $ "getSigParms: " ++ s
```
m0dev76: Again, I'm not expecting users to do full structural validation of transactions. I don't think this is a programming error. Perhaps we disagree on this point?

tests/Network/Haskoin/Script/Tests.hs
```diff
@@ -33,8 +33,6 @@ tests =
     , testGroup "Script Parser"
         [ testProperty "decode . encode OP_1 .. OP_16" testScriptOpInt
         , testProperty "decode . encode ScriptOutput" testScriptOutput
-        , testProperty "decode . encode ScriptInput" testScriptInput
```
m0dev76, line 36: This test has to remain. It verifies decode . encode of ScriptInput. The test needs to be adapted to the new library.

tests/Network/Haskoin/Transaction/Arbitrary.hs
```diff
@@ -139,7 +139,7 @@ genPayTo = do
 -- Generates data for signing a PKHash transaction
 instance Arbitrary PKHashSigTemplate where
     arbitrary = do
-        inC   <- choose (0,5)
```
m0dev76, line 142: Why only generate valid test data? We also want some invalid test data to make sure our libraries behave correctly. Having transactions with no inputs is a good way to verify the behaviour of the signing code in such cases.

m0dev76:
As a general comment, I prefer the solution I have here for ScriptInput an decodeInput:
https://github.com/plaprade/haskoin/blob/unify-scriptinput/Network/Haskoin/Script/Parser.hs
Otherwise, I think the way you split up the functions is nice. However, we fundamentally disagree on the usage of errors in this code.

jprupp:
I have added some code to address plaprade's concerns. I still think strongly against having a separate data type for P2SH. P2SH is not so special.

One thing not commented here is the asymmetry of the output of signing functions for transaction and input. One solution would be to not output transaction status in the signing function, but I agree that it is convenient for the user of the high level function. Another solution would be for the signInput code to sign with multiple keys simultaneously, so that the aggregate of multiple signing operations on multisig inputs can be addressed. I like the first solution better: to separate completion validation from transaction signing, especially since I expose sigStatusTx precisely for that. Removing signature completion status from the transaction signature path would simplify things too. I think I am going to work on this.

jprupp:
I decided to go with plaprade's parser. I'm changing my code to use that.

jprupp:
Pull request #50 supersedes this one.

PR #52, opened 2014-06-29
Fixing dependencies to work with the new script evaluator

m0dev76:
cabal install --enable-tests didn't find a valid solution for dependencies.


jprupp:
ACK

PR #56, opened 2014-06-30
support for base-4.7

OttoAllmendinger:
I'm using Arch Linux where base-4.7 is default now, Haskoin runs fine with it. I think this can be changed safely.


jprupp:
ACK

I'd put "base >= 4.6 && base < 5".

m0dev76:
Yes I agree, let's go for base < 5

jprupp:
Superseded by #58

PR #57, opened 2014-06-30
Fix script tests

OttoAllmendinger:
The test suite `script_valid.json` passes again

Also adds `script_invalid.json` file testing rejection of invalid scripts


jprupp:
ACK

PR #60, opened 2014-06-30
Simplify Mnemonic and remove dependency on ICU

jprupp:
Mnemonic simplification:
- Does not support non-ASCII mnemonics
- Does not depend on ICU library
- Only supports English word list
- Uses String instead of Text
- Uses Vector instead of a list for dictionary
- Has embedded Map to quickly find words in dictionary


sha49:
much ack

m0dev76:
You should update the readme as this line is not needed anymore:
`sudo apt-get install libicu-dev`

jprupp:
Fixed readme in pull request #61.

sha49:
ack

PR #64, opened 2014-07-01
Removing the ScriptOutput requirement for decoding a ScriptInput

m0dev76:


jprupp:
ACK

That's what I had in mind indeed!

PR #71, opened 2014-07-07
Fix of PayToSH logic in Evaluator.hs

lornap:
A couple of bug fixes in the p2sh logic.
- Added more checking on whether the redeemScript should be evaluated
  (no empty stack, scriptPubKey had success)
- Changed the starting program state to be what is required ( need the scriptSig
  ending stack minus the redeemScript, and hashOps needs to be the redeemScript)

Admittedly, this fix is a bit of a kludge.  Some clean-up or refactoring of the area is probably a good idea, but I wanted to get the fix in.


m0dev76:
Thanks for the fix. We should probably also add some unit tests of programs containing real signatures / redeem scripts that would fail/pass in such cases.

jprupp:
ACK

sha49:
ACK

PR #72, opened 2014-07-07
Complete rewrite of JSON-RPC

jprupp:
A very different JSON-RPC interface based on type classes.


tests/data/invalid.json
```diff
@@ -1,6 +0,0 @@
-{"jsonrpc": "2.0", "method": "foobar, "params": "bar", "baz]
-{"jsonrpc": "2.0", "method": 1, "params": "bar"}
-{"jsonrpc": "2.0", "method"
-[]
-[1,2,3]
-{"foo": "boo"}
```
m0dev76, line 6: Why remove these tests? Are they no longer relevant?

jprupp, line 6: Yes, these tests are no longer relevant.

m0dev76, line 6: How so? They seem to be relevant for JSON-RPC. Do you have equivalent tests for these somewhere? Maybe I missed them

jprupp, line 6: The way the library is written now makes these tests redundant. I can add them if you wish, but I don't see much value in them anymore.

m0dev76:
Are they redundant because other tests check the same thing? Otherwise you usually want to keep tests because they give you guarantees that you don't break something while refactoring in the future. Removing tests because the library passes them doesn't make much sense :) Maybe you have a good reason to remove them, but your replies have been a bit vague.

jprupp:
I used to have generic data types where I could accommodate a partially-parsed Aeson Value that was a structurally valid JSON-RPC request or response. It made sense to attempt to parse anything to these data types, including invalid data.

The current iteration of the code declares functions that take an element of type Value, and outputs data that belongs to user-defined data types inside the Aeson Parser monad. This removal of the intermediate data type makes the type system deal with invalid data. The parser just returns failure upon being unable to create Haskell data, it is the only thing that it can do.

That being said, the tests may still be performed, and I'm adding them back, because they may still be used to detect exceptions that cannot be handled by the Parser monad.

Network/Haskoin/Stratum/RPC.hs
```diff
@@ -0,0 +1,118 @@
...
+, encodeNotif
+, parseNotifParams
+, parseNotif
+  -- ** Encoding
+) where
+
+import Control.Applicative ((<$>))
+import Data.Aeson.Types
+import Data.Text (Text)
+import Data.IntMap.Strict (IntMap)
```
m0dev76:
```
Network/Haskoin/Stratum/RPC.hs:31:1: Warning:
    The import of `Data.IntMap.Strict' is redundant
      except perhaps to import instances from `Data.IntMap.Strict'
    To import instances alone, use: import Data.IntMap.Strict()
```

jprupp: Yup, noticed it. I will update commit soon.

m0dev76:
ACK

PR #74, opened 2014-07-16
Util: improve bsToHex performance

OttoAllmendinger:
The old routine is very slow for larger inputs, something O(n^2):

```
> length $ bsToHex $ BS.pack $ take 3000 (repeat 0x62)
6000
(1.01 secs, 662956664 bytes)

> length $ bsToHex $ BS.pack $ take 6000 (repeat 0x62)
12000
(4.89 secs, 2993737936 bytes)
```

The new method is considerably faster:

```
> length $ bsToHex $ BS.pack $ take 6000000 (repeat 0x62)
12000000
(0.72 secs, 653576448 bytes)
```

### DIFF
--- a/Network/Haskoin/Util.hs
+++ b/Network/Haskoin/Util.hs
@@ -97,6 +97,10 @@ import qualified Data.ByteString as BS
     , foldl'
     , null
     )
+import qualified Data.ByteString.Builder as BSB
+    ( toLazyByteString
+    , byteStringHex
+    )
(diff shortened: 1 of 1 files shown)


jprupp:
ACK

I wanted to do this one year ago, but the version of ByteString we depended on back then did not have the functionality. :+1:

m0dev76:
That's a nice improvement. Thanks!

PR #75, opened 2014-07-16
Script: OP_PUSHDATA can have zero-length payload

OttoAllmendinger:
There were some byte sequences that could be decoded into Script, but could not
be encoded back into byte sequences again.

Also fixed `opPushData BS.empty`

Add tests asserting `(encode (decode bytes)) == bytes` for canonic tests cases

Should now behave identically to the reference implementation:
https://github.com/bitcoin/bitcoin/blob/0.9.2/src/script.h#L372


jprupp:
ACK

PR #76, opened 2014-07-18
Fix script

OttoAllmendinger:
some edge cases found with QuickCheck


OttoAllmendinger:
there may be more to come

OttoAllmendinger:
This should be it for now. 

I have a suspicion that there are some other bugs, but I have to look at it again another time.

m0dev76:
Nice fixes, thanks! ACK on my end.

jprupp:
ACK

sha49:
ACK

Thanks @OttoAllmendinger

PR #79, opened 2014-07-21
Json instances

sha49:
FromJson / ToJson instances for Coin, ScriptOutput, Tx and Outpoint added


m0dev76:
Great! However, it would be nice to have unit tests fore decode . encode = id

sha49:
I added two more Json instances and a test group for all the (ToJson, FromJson) instances

m0dev76:
ACK

PR #84, opened 2014-07-25
Add Network.Haskoin.Internals to export haskoin internals

np:
This enable libraries and tools to re-use the great code available here.
The rule is that no guarantee on the stability of such module.


m0dev76:
That's a great idea. Thanks for sending in a pull request. ACK

jprupp:
ACK, this is very cool!

sha49:
ack

PR #93, opened 2014-08-09
transaction signing functions now return a complete flag corresponding t...

m0dev76:
...o verifyStdTx


jprupp:
I'm OK with this. ACK.

PR #104, opened 2014-08-18
mtl for NixOS

jprupp:
NixOS maintainers require us to depend on a slightly older version of mtl. We know it works with the older version because we bumped the version number recently, and I believe we have not changed anything in the code that will break compatibility with the older version.


sha49:
ack

PR #110, opened 2014-09-23
Fixed tests on 32-bit Linux

jprupp:
Various fixes for 32-bit Linux tests.


m0dev76:
Good job on getting those bugs fixed!

PR #115, opened 2014-10-30
HeaderChain

m0dev76:
This patch adds a HeaderChain implementation to Haskoin. It is a headers-first chain implementation with optional fast catchup. It can validate the proof of work of block headers and is meant to be used together with an SPV client.


jprupp:
ACK

I gave only cursory check, but I like the design.  I did not do a detailed audit of the algorithms themselves.

PR #116, opened 2014-11-11
Peer, PeerManager and SPVNode

m0dev76:
This pull requests adds a generic Peer and PeerManager module for connecting with other Bitcoin nodes. It also implements a simple SPV node that can deliver transactions and merkle blocks to a user application.


jprupp:
ACK

sha49:
ack

PR #120, opened 2014-11-29
Replaced deprecated ErrorT with ExceptT in Evaluator.hs

lornap:
Simple replacement of ErrorT with ExceptT.  Almost a trivial
replacement.


jprupp:
ACK

Missed you lornap!

lornap:
Been swamped.  Sorry about that.

m0dev76:
Thanks for the fix!

PR #123, opened 2014-12-04
Bug fix of potentially dangerous "head" call.

lornap:
Code to ensure the call to head is done safely when the stack is
empty.


jprupp:
ACK

PR #124, opened 2014-12-04
Bug fix: Failure handling in P2SH script decoding

lornap:
Preventing a failure to decode from crashing the application.


jprupp:
ACK

PR #128, opened 2014-12-13
No GADTs necessary

jprupp:


m0dev76:
ACK, I guess phantom types are a bit too much for this simple problem.

PR #138, opened 2015-01-11
Do not recompile haskoin libraries twice when running tests

m0dev76:
No functional changes. Mostly streamlining the cabal build process.


jprupp:
ACK

PR #139, opened 2015-01-13
Getting rid of DerivNonPrime

jprupp:


m0dev76:
Can you also uncomment the JSON test for DerivPath and see if the tests pass?

Network/Haskoin/Test/Crypto.hs
```diff
@@ -211,12 +213,23 @@ instance Arbitrary ArbitraryXPubKey where
...
+        ArbitraryDerivPath <$> elements 
+            [ DerivPrv path
+            , DerivPub path
+            ]
+
+instance Arbitrary ArbitraryDeriv where
+    arbitrary = do
+        w <- arbitrary
+        let i = w .&. 0x7fffffff
+            b = w ==  0x80000000
```
m0dev76: I don't think this line is correct.

m0dev76:
ACK

PR #141, opened 2015-01-20
Phantom types for Network

jprupp:


jprupp:
Gravitated to solution in pull request #143

PR #143, opened 2015-01-26
Possible to change network constants at runtime

jprupp:


m0dev76:
I think this version is much cleaner than the previous one using phantom types. Here are a few comments. 

I think you should provide functions for switching the network instead of having users call `writeIORef networkRef testnet3` and expose the IORef to the user. Perhaps a function like `setNetwork :: Network -> IO ()`.

Adding rotate to BigWord is also a good idea as it's part of the minimal complete definition for Bits. If you could add a property test for rotate in this file that would be great: https://github.com/haskoin/haskoin/blob/master/tests/Network/Haskoin/Crypto/BigWord/Tests.hs. The test checks that BigWord Mod32 is equivalent to Word32.

m0dev76:
You also need to use this pragma for all functions that use unsafePerformIO: `{-# NOINLINE foo #-}`. This makes sure that the function doesn't get automatically inlined by GHC as an optimization. Although there are no side effects in our particular case, using unsafePerformIO breaks referential transparency so we can't let GHC inline the functions as it pleases.

m0dev76:
ACK

PR #146, opened 2015-03-03
LTS Haskell

jprupp:


m0dev76:
If we use LTS Haskell, what is the advantage of keeping version ranges for modules in the haskoin.cabal file? Souldn't we also align those versions to the LTS versions? That way, people that don't use LTS Haskell would still be guaranteed a correct build and a predictable and tested build.

jprupp:
I don’t think so. Policy should not be in our cabal file.  The Cabal file in the application merely states what is compatible at the API level with our repository, while the config.cabal from Stackage is a policy.  It states exact versions.  We may update the config.cabal at any time, and it does not necessarily mean that we will update our cabal file.  A user that needs to compile haskoin on a non-Stackage system may decide to remove config.cabal for that.

m0dev76:
OK, sounds good then. ACK

PR #149, opened 2015-03-13
Reworking extended keys and derivations

m0dev76:
Instead of working in the Maybe monad for derivation failures that will never happen in practice (as likely as a hash collision), I simply throw a DerivationException. 

I removed the NormalizedKeys interface which was a bit bloated and unfriendly. I provide simpler address derivation functions in ExtendedKeys. 

I also fixed a few warnings on GHC 7.8.4 and aligned tests.


m0dev76:
I've rewritten the custom path derivation code. It is now more typesafe and provides very convenient Haskell constructors to build derivation paths.

jprupp:
ACK (I did not review much)

sha49:
I will take a look and see if this can be done in purescript :)

m0dev76:
I've added some helper functions to derive addresses from a custom parent SoftPath. This is useful for deriving internal or external addresses for example.

Network/Haskoin/Crypto/ExtendedKeys.hs
```diff
@@ -372,7 +332,252 @@ instance Binary XPubKey where
...
+    go f p = case p of
+        next :/ i -> go (f . flip pubSubKey i) next 
+        _         -> f
+
+-- | Derive a key from a derivation path and return either a private or public
+-- key depending on the initial derivation constructor. If you parsed a string
+-- as m/ you will get a private key and if you parsed a string as M/ you will
+-- get a public key. If you used the neutral derivation constructor `Deriv`, a
+-- private key will be returned. 
+derivePathE :: DerivPathI t -> XPrvKey -> Either XPubKey XPrvKey
```
jprupp, line 543: I love this function.

Network/Haskoin/Node/SpvBlockChain.hs
```diff
@@ -247,12 +247,20 @@ headerSync resource locType hStopM = do
...
-processMerkleBlocks did [] = do
-    $(logError) $ format $ "Got a completed merkle job with an empty merkle list"
-    modify $ \s -> s{ merkleId = Nothing }
-    -- We continue the merkle download because we didn't have any transactions
-    continueMerkleDownload
-    checkSynced
+processMerkleBlocks did [] = gets merkleId >>= \mid -> case mid of
+    Just (expId, _) -> if did == expId
+        then do
+            $(logError) $ format $
```
jprupp, line 253: Is this really an error condition?

m0dev76, line 253: We should not receive an empty merkle block job from a peer as we only create non-empty merkle download jobs for peers. I think we should never hit this computation branch but I log an error here just in case we do get here in an unexpected way.

jprupp:
ACK

PR #150, opened 2015-03-23
Request a full block locator on a block tickle

m0dev76:
On a block tickle, we have to issue a full block locator. Otherwise, on a reorg, the remote peer will send us everything again from the genesis.


jprupp:
ACK

PR #153, opened 2015-04-05
Add toMixed function and return more information for address derivations

m0dev76:
There are 2 changes in this pull request.
- Add a `toMixed` function that transforms an arbitrary `DerivPathI t` into a `DerivPath` (mixed path). This is always possible and will always type check as a mixed path can contain any type of derivations.
- Address derivation functions now also return the associated redeem script for multisig addresses or the public key for regular addresses. This is helpful further up in wallet libraries to prevent duplication of work.


jprupp:
ACK. Very useful.

PR #155, opened 2015-04-20
Add an appendPath function to append mixed derivation paths together

m0dev76:


jprupp:
ACK

jprupp:
I believe there are more details to eventually consider to choose coins than the value. The age of coins is important too in order to drive transaction fees down. It is also preferrable to spend confirmed coins than mempool ones. Perhaps enriching the Coin class to include this would be a good idea.

m0dev76:
Sure, having more policies for choosing coins is a good idea. However, I would only add new functions to the Coin class as we need them (together with the related code). I don't want to guess what functions we might need later as I might not know. The only requirement I currently have for coin selection is the value of a coin, which is why it is the only function required in the class.

jprupp:
ACK

PR #157, opened 2015-05-02
No Stratum

jprupp:
I do not plan to upgrade Stratum protocol support so that it is up-to-par with SPV. I believe we sholud focus on SPV only. I am moving Stratum to a separate package.


m0dev76:
ACK

PR #158, opened 2015-05-11
Implementing Show and Read instances for DerivPath

m0dev76:


jprupp:
Hey, what about those empty strings for Deriv? That kind of derivation will look like "/0/3", no starting character. If we make it 'm', I think it breaks symmetry. Is that the reason why it is left empty?

m0dev76:
If you put an m/ in front of a Deriv, it will read as a DerivPrv and break symmetry. The reason why I have Deriv as well as DerivPrv and DerivPub is to be able to only represent a derivation path without having to add information of whether it is a private or public derivation. In the wallet case, the same derivation can be used to create public keys (for addresses) and private keys (for spending the coins). So adding an m/ all the time is confusing.

jprupp:
ACK

PR #160, opened 2015-06-04
Add conduit-based coin selection functions and a continue flag

m0dev76:


jprupp:
A very nice solution. Kudos. The faucet will thank you.

PR #161, opened 2015-06-05
Fix a bug while connecting child nodes in the HeaderTree

m0dev76:
Also added some extra debug-level logs in the node.


jprupp:
ACK

PR #163, opened 2015-06-07
Support recipient pays fees

jprupp:
These changes allow for recipients to pay for transaction fees by having the fees deducted from the amount payable.

### DIFF
--- a/Network/Haskoin/Transaction/Builder.hs
+++ b/Network/Haskoin/Transaction/Builder.hs
@@ -26,7 +26,8 @@ import Control.Monad.Trans.Either (EitherT, left)
 import Control.Monad.Identity (runIdentity)
 import Control.DeepSeq (NFData, rnf)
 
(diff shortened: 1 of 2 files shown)


m0dev76:
I'm not sure I agree with the way you have implemented this functionality. If you want the recipient to pay the fee, then you don't want to pay a fee yourself, so you should set fee = 0 in `chooseCoins`. So you do not need to add an extra `rcptFee` argument in all those functions and you don't need to change the `greedyAddSink` function either.

Now, you do want to know how much fee the recipient has to pay in order for the transaction to be valid. This is trivial: you can use one of the guessing functions `getFee` or `getMSFee` which will give you an upper bound on the fee that you have to pay. This is what `greedyAddSink` uses anyway so you won't get a more precise result either way. So, I think that we do not need this whole pull request to implement the functionality that you require. You probably only need to export `getFee` and `getMSFee` from Transaction.hs to have access to them from the haskoin-wallet project.

jprupp:
LOL, you are absolutely right.

PR #165, opened 2015-06-14
Support transaction retransmission

jprupp:


haskoin.cabal
```diff
@@ -125,6 +126,7 @@ library
                  , stm-conduit              >= 2.5          && < 2.6
                  , text                     >= 0.11         && < 1.3
                  , time                     >= 1.4          && < 1.5
+                 , unordered-containers     >= 0.2          && < 0.3
```
m0dev76: Is there a benefit to using Data.HashMap.Strict over Data.Map.Strict? These 2 dependencies are not required by the later.

Network/Haskoin/Node/Mempool.hs
```diff
@@ -154,6 +167,29 @@ processTxInv pid tids = do
...
+             -> TxHash
+             -> StateT MempoolSession m ()
+processGetTx pid tid = do
+    $(logDebug) $ format $ unwords
+        [ "Requesting transaction", encodeTxHashLE tid, "from wallet." ]
+    modify $
+        \s -> s{ txPeerMap = insertWith (flip (++)) tid [pid] (txPeerMap s) }
+    sendWallet $ WalletGetTx tid
+
+processSendTx :: (Applicative m, MonadIO m, MonadLogger m)
```
m0dev76: I have nothing against Applicative per se, but to stay consistent with the rest of the code in the node, I would suggest using liftM as we have that already with the Monad constraints.

m0dev76:
I like the pull request. However, in addition the few comments I have posted on the code, I was wondering if it is reasonable to re-broadcast all pending transactions for every new block. If we manage a very large wallet in which we might have transactions that are stuck for various reasons, we might start consuming a lot of resources just re-broadcasting transactions.

I think it would be reasonable to have some kind of exponential backoff for re-broadcasting transactions, perhaps a time-based backoff that is checked everytime a new block comes in. That would prevent us from accumulating large amount of broadcasts in some situations.

Network/Haskoin/Node/Mempool.hs
```diff
@@ -154,6 +166,29 @@ processTxInv pid tids = do
...
 
+processGetTx :: (MonadIO m, MonadLogger m)
+             => PeerId
+             -> TxHash
+             -> StateT MempoolSession m ()
+processGetTx pid tid = do
+    $(logDebug) $ format $ unwords
+        [ "Requesting transaction", encodeTxHashLE tid, "from wallet." ]
+    modify $
+        \s -> s{ txPeerMap = insertWith (flip (++)) tid [pid] (txPeerMap s) }
```
m0dev76: What if the same PeerId sends us the same TxHash twice ? Should we nub the list ?

jprupp: Yes, we should. I will do so.

jprupp:
Since I have addressed all concerns about this pull request. I will proceed to merge it.

PR #167, opened 2015-06-19
Mempool synchronization

jprupp:
Synchronize mempool against all peers after blockchain sync done, only once.


Network/Haskoin/Node/Mempool.hs
```diff
@@ -94,6 +94,8 @@ data MempoolSession = MempoolSession
       -- Map of transactions to peer ids to respond to GetData requests
       -- from peers.
     , txPeerMap :: !(Map TxHash [PeerId])
+      -- If mempool has been synced at least once, do not do it again
+    , mempoolSynced :: !Bool
```
m0dev76: I think your new field here is exactly the same as the `nodeSynced` field. Maybe you can reuse `nodeSynced`?

m0dev76:
ACK

PR #172, opened 2015-07-07
Send regular keepalives

jprupp:


m0dev76:
If I understand correctly, the MVar is used to cancel/reset the ping timeout if we received a message from the peer. Do we really need that or is it sufficient to just send a ping at a regular interval regardless of whether we received anything from the peer?

jprupp:
As discussed offline, the MVar is used as a semaphore so the peer knows to disconnect when no data has arrived for more than 90 seconds. If we were to send pings at regular intervals, we would still be at the mercy of arbitrary OS TCP timeouts. Most network applications control their own TCP retransmission and timeout settings. I set Haskoin’s to be more aggressive than those of bitcoind, so that it works better on a mobile/laptop environment, while not negatively impacting running on a server.

m0dev76:
OK, nice patch. ACK

PR #173, opened 2015-07-19
Reject fix

m0dev76:
This pull request also adds simple node status logging and a fix to the nodes network height tracking.


jprupp:
ACK

m0dev76:
I've added height indices in the header tree so that we can build block locators quickly without having to scan the entire chain. I've also removed some redundant information that was being kept around in memory while downloaded merkles blocks.

PR #174, opened 2015-07-28
Move node actors to a separate haskoin-node library

m0dev76:
I will move the bitcoin node packages into their own module to keep dependencies of the pure haskoin code simple.


jprupp:
ACK

Issue #175, opened 2015-08-01
Upgrade deepseq to >=1.4?


nuttycom:
I am attempting to write an implementation of the BIP-70 Payment Protocol, and would like to take advantage of some of the Haskoin types in my implementation. In the process, I've discovered that the protobuf library (which I need for serialization of the Payment Protocol messages) relies on a more recent version of deepseq than haskoin is using. Would you accept a pull request that upgrades the deepseq version?

jprupp:
I have upgraded this library to use LTS Haskell 3.4, which includes deepseq 1.4.1.1. I’ll close this ticket.

PR #176, opened 2015-08-01
Upgrade deepseq to >=1.4

nuttycom:
See #175

### DIFF
--- a/Network/Haskoin/Node/Actors/PeerManager.hs
+++ b/Network/Haskoin/Node/Actors/PeerManager.hs
@@ -33,6 +33,7 @@ import qualified Data.Map as M
     , null, fromAscListWith, toAscList
     , mapMaybe
     )
+import GHC.Generics (Generic)
 
 import Network.Haskoin.Util
 import Network.Haskoin.Block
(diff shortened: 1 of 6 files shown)


xenog:
ACK

jprupp:
We use Stack and LTS Haskell to compile Haskoin. We can push the upper bound for deepseq, but not the lower bound, since deepseq 1.4 is not yet in the latest LTS Haskell.

m0dev76:
We just merged in a pull request that moved most of the node to the haskoin-node project to simplify dependencies in haskoin. Some of your changes would now be in haskoin-node. 

Additionally, I think that if we decide to provide Generic instances for some types in the haskoin library, we should provide them for all the useful types that are exposed. It would make sense from a consistency point of view. That is at least what we try to do for other instances such as Json and NFData even though we might be missing quite a few.

jprupp:
Hello @nuttycom,

In order to merge this pull request, we would like to get Generic instances for all exposed data types, instead of just a few of them, and also rebase against the latest master. Wolud you do this?

Cheers!

nuttycom:
Sure, I can do this. It may take me a few days to get to it. Thanks for the
LTS upgrade!

jprupp:
Send new pull request when done. I'll close this one.

PR #186, opened 2015-10-14
Fix truncating String to ByteString conversion

### LINKED ISSUE #184: Base58 malleability vulnerability due to ByteString conversion
The function `stringToBS` is implemented as `Data.ByteString.Char8.pack`, which truncates characters longer than 8 bits.  This means that a `String` may have unexpected `Base58` encoding.

jprupp:
Fixes #184.

### DIFF
--- a/Network/Haskoin/Crypto/Base58.hs
+++ b/Network/Haskoin/Crypto/Base58.hs
@@ -27,7 +27,6 @@ import Data.Aeson
 
 import Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
-import qualified Data.ByteString.Char8 as C
 import qualified Data.Text as T
 import Data.Text.Encoding (encodeUtf8)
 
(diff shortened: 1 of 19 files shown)


m0dev76:
I've been phasing out the use of `bsToString` and `stringToBS` because they were simply aliases to Char8.pack and Char8.unpack. I have been using the later versions in most of the newer code, especially the wallet. 

I am unsure if these functions are used anywhere in the network protocol path and if anything would break if we now decode results from the network as UTF8. This needs to be double checked.

m0dev76:
Actually, I would rather advocate for removing those two functions entirely and use the appropriate conversions where necessary. We can also keep those two functions (the UTF8 variants) but renamed them such that their function is clearer. This will force us to go through the code and change all the instances of `bsToString` and `stringToBS`.

jprupp:
I agree.  I have a ticket open for that.  I will do it.  For now, since this is a potential security issue, I'll merge this quick dirty fix, if you ack it.

jprupp:
This request will be superseded.

PR #189, opened 2015-10-19
Add testcases hard-follows-soft paths (find_some_failcases_for_xkey_parsePath)

tphyahoo:
This follows branch extended_key_test_vector_documentation, which still has tests passing.

With current branch, tests fail.

I put this into its own branch, in case I misunderstood something about bip32 spec and these failure cases are in fact not legitimate.  In that case, extended_key_test_vector_documentation should still be pulled, but current branch should be rejected.

In any case, the issue is that

everything that depends on paths of form  m/0'/1/2' now fails.

```
I think the pattern that breaks it is hardened, followed by non-hardened, followed by hardened.

parsePath fails here.
```


tphyahoo:
fixes #192

jprupp:
This is on purpose. We don't want to support hardened paths following non-hardened paths. That is why @plaprade wrote that complex type-based solution. It actually makes the type system take an active role in guaranteeing that bad use cases are not possible.

tphyahoo:
Why not allow hard-follows-soft paths, if it's allowed under the original bip32 spec test vectors. This is library code after all.

If you want to disallow hard-follows-soft for haskoin-wallet and haskoin-sx, etc, you could have a different type safe way of dis-allowing those combinations.

I think what you want to do is differentiate between hard and soft paths in a type safe way, so that a pub->pub derivation on a hard path gets rejected at compile time.  But that's not the same thing.

These hard-follows-soft paths I added to tests are valid hardened paths.

tphyahoo:
In the new stuff in 

rework_ExtendedKeys

branch, hard keys are 

```
data XKeyHardPath = 
                      (:|/) XKeyHardIndex XKeySoftPath
                    | (:/|) XKeySoftIndex XKeyHardPath
                    | (:||) XKeyHardIndex XKeyHardPath
  deriving (Read,Show,Eq,Ord)
```

We then define 

```
instance XPrvKeyEndoDerivable XKeyHardPath
instance XPrvKeyEndoDerivable XKeySoftPath
instance XPubKeyEndoDerivable XKeySoftPath
```

but not 

```
instance XPubKeyEndoDerivable XKeyHardPath
```

Perhaps you want to forbid hard-follows-soft for certain applications. Well, all you need to do then is hide the original data declaration and replace with

```
data XKeyHardPath = 
                      (:|/) XKeyHardIndex XKeySoftPath                   
                    | (:||) XKeyHardIndex XKeyHardPath
  deriving (Read,Show,Eq,Ord)
```

jprupp:
Hey @tphyahoo,

First, sorry for responding this late, it has been a crazy pair of weeks. I really like the work you have done in your [rework branch](https://github.com/tphyahoo/haskoin/tree/rework_ExtendedKeys). I would like to have all of your proposed changes for BIP-32 squashed into a single commit rebased against the the secp256k1 branch, and then a single pull request be sent for that. We will just fast-forward the master branch then to your commit, and the pull request will be marked as merged (yes, GitHub can detect Linus-style merges).

This is all pending @plaprade approval, but I don't expect any objections from him. Your code is much better than what was there before.

Feel free to close this and related requests and then open a single one with the whole BIP-32 rework. I promise I will review it faster and get @plaprade to do so too and get it merged.

Greetings

jprupp:
I forgot, yes, add strictness annotations to your data types. It is good for performance not to create unnecessary thunks.

tphyahoo:
Good to hear you liked the work.

I added strictness annotations.

I am used github flow for collaborating.

https://guides.github.com/introduction/flow/

With github flow, you would get your secp256k1 branch blessed into master.  Then I would merge master, and submit my work on top of that.  Or it could be the other way (mine first into master, then secp256k1 off that), but the point is that major features are incorporated into master as the coordination point.  

Would that be ok? Otherwise what is the point of my basing rework_ExtendedKeys branch off secp256k1? Is it just to make it easier for you to review? (Just trying to understand.)

How far from merge to master do you think secp256k1 branch is?

jprupp:
Branch secp256k1 is very close to becoming master very soon, that is why I recommend to base your work on it. It is next-in-line to become merged.

The style of merge that we have adopted from Haskoin is somewhat of a hybrid between the GitHub flow and the Linus flow. We like to keep the master branch free from merging commits, and we like to make linear (in time) commits with specific features from other branches, be them in the same repository or in clones. After discussing a pull request and deciding that will become merged, we like to squash all commits into one, and then fast-forward the master branch with that single commit. This has the effect of leaving the history of the repository quite clean, and linear.

tphyahoo:
It seems that if I do as you say it would violate

https://www.atlassian.com/git/tutorials/merging-vs-rebasing/the-golden-rule-of-rebasing/

since rework_ExtendedKeys is a public branch (albeit not a huge number of eyes on it).

I guess this is a soft rule, or are there extenuating circumstances in this case I'm not aware of?

From the Pro Git Book:

"In general the way to get the best of both worlds is to rebase local changes you’ve made but haven’t shared yet before you push them in order to clean up your story, but never rebase anything you’ve pushed somewhere."

"http://www.git-scm.com/book/en/v2/Git-Branching-Rebasing#Rebase-vs.-Merge"

tphyahoo:
In light of this, I will probably either

a) do nothing till secp256k1 gets into master or
b) merge (but not rebase) secp256k1 into rework_ExtendedKeys, in anticipation of making future merge to master easier, without having to deal with messiness of rebase.

Is this ok, or do you still really prefer a rebase?

tphyahoo:
superceded by #200.

tphyahoo:
reopening since #200 is no-pull.

tphyahoo:
reclosing, superceded by #202

PR #197, opened 2015-11-21
Merge haskoin, haskoin-node and haskoin-wallet

m0dev76:


jprupp:
You forgot to rename the package `haskoin` to `haskoin-core`. The contents of `haskoin.cabal` as well as its filename must change.

Issue #192, opened 2015-10-26
hard-follows-soft paths for bip32 wallets are not properly tested.


tphyahoo:
Paths like m/0'/1/2' and m/0/2147483647' are not being tested properly in current test suite.

"hard follows soft" paths are legitimate paths for privkey derivations, they are even included in official bip32 test vector.

tphyahoo:
this is fixed by PR https://github.com/haskoin/haskoin/pull/189

tphyahoo:
The Bip32 code is basically done at this point.

It probably still needs some cleanup before merge.

In particular, I'd like all the packages that depend on haskoin-core compiling cleanly.

I think some review would also be a good idea.

I have changed some of the Bip32 behavior, because I think the old behavior was bugged.

In particular, in

https://github.com/tphyahoo/haskoin/blob/rework_ExtendedKeys_take3/tests/Network/Haskoin/Crypto/ExtendedKeys/Units.hs

please see new vectors for 

```
deriveHardPrvPathVectors :: [(XPrvKey, String, Maybe XPrvKey)]
derivePathVectors :: [(Bip32XKey, String, Maybe Bip32XKey)]
derivePubPathVectors :: [(XPubKey, String, Maybe XPubKey)]
```

I hope to have something polished and reviewed enough for merge very soon.

tphyahoo:
There seems to be some directory shuffling going on, so better to use "take3" than take2" branch, which got polluted by the shuffle.

I updated above comment to reflect this.

tphyahoo:
experimental branch

https://github.com/tphyahoo/haskoin/tree/rework_ExtendedKeys_take4

has all tests (including haskoin-wallet) tests passing.

I think this isn't quite ready for a PR because it should be structured as

-- patch haskoin
-- patch haskoin-wallet

and then pull.

And I need to do the rebase stuff we discussed earlier.

It might be worth having a diff + look just to see if anything seems suspicious neverthelesss.

It is mergable to master as is.

Issue #196, opened 2015-11-15
bip32 public key derivations (starting with capital M) aren't handled correctly


tphyahoo:
From Network.Haskoin.Crypto.ExtendedKeys.Units.derivePrvPathVectors

```
, ( xprv, "m/8'", hardSubKey xprv $ XKeyHardIndex 8 )
, ( xprv, "M/8'", hardSubKey xprv $ XKeyHardIndex 8 ) 
```

this test passes if behavior is identical for M as for m.

I believe the convention is for M to derive a public key. It's a bit sketchy in the bip32 spec but

https://github.com/GemHQ/money-tree has this, and the current repo also has in 

Network.Haskoin.Crypto.ExtendedKeys.parrsePath: 

```
pEnd = case x of
    ""  -> Just Deriv
    "m" -> Just DerivPrv
    "M" -> Just DerivPub
    _   -> Nothing
```

 So I think the above test needs fixing.

Right?

tphyahoo:
My solution for the above issue was to replace derivePrvPathVectors with deriveHardPrvPathVectors.

See 

https://github.com/tphyahoo/haskoin/commit/b6574dc5961c820bf3147adc4325bb514e2e4e54#diff-b4a7936eaf19e08d409c64c9fb5a6a4a

and ctrl-f for these functions in the large diff.

Test strings must now parse as hard path. Soft paths, or DerivPath parsable strings are no longer allowed.

I think library users will probably usually want to use deriveBip32Path (which replaces derivePathE, which I left as a synonym for now but would like to deprecate), which is the most flexible method. For deriveBip32Path, you can start with either a public or private key. You use new data type Bip32XKey which allows either pub or priv as start point. And since DerivPath transformations can point to either private key (with lower case m) or public key (with upper case M), the result type is again Bip32XKey. Well actually, the result type is Maybe Bip32XKey to allow failures in the case of impossible derivations. 

Impossible derivations include: m/1/2/3 starting with pub key.  Or M/1'/2'/3' (a hard path) starting with pub key.

(Perhaps this should be added to haddock.)

For new behavior of derivePathE, ctrl-f derivePathVectors in above linked commit.

tphyahoo:
This can be closed as soon as #200 is accepted to haskoin-core.

tphyahoo:
In light of 

https://github.com/haskoin/haskoin/pull/243 and https://github.com/haskoin/haskoin/pull/242

[...]

PR #193, opened 2015-10-26
clean up testing code for bip32 wallets (tests_chain_in_vector)

tphyahoo:
This would be an easy merge after 

find_some_failcases_for_xkey_parsePath

is merged into master.

It doesn't add new functionality or change behavior, just simplifies  testing code.

The biggest improvement is it stops intertwingling the logic for the derivation path in testing vector, and puts everything into one places for testing, pretty much following the bip32 spec.

PR #188, opened 2015-10-18
Improve tests for xpubkeys derivation (extended_key_test_vector_documentation)

tphyahoo:
factor tests and reduce amount of code.

I believe I may have also found some bugs in the test suite, but I will put that into a separate PR.

As of this PR, all tests still pass.

PR #200, opened 2015-12-03
BIP-32 rewriting experiment

### LINKED ISSUE #191: bip32 xkey parsing is overly complicated
I am overhauling the bip32 xpubkey/xprivkey parsing and derivation machinery

This work began when I attempted to fix #192, and I noticed the current implementation is overly complex and therefore hard to fix.

I am working on simplifying the bip32 xkey handling machinery, in branch

https://github.com/tphyahoo/haskoin/commits/rework_ExtendedKeys

Hopefully I should have something ready to pull soon.

tphyahoo:
I have a merge candidate for haskoin/haskoin that is 

-- squashed (1 commit)
-- all tests pass for haskoin and sub-repos (haskoin-wallet required additional fix for compile and pass)

The downside is

-- it's a lot to review: 14 files, and 40 pages printed out.
-- it may not be properly staged. It upgrades haskoin and haskoin-wallet simultaneously, and I'm not sure if that's what is wanted or if this is acceptable. The alternative being to stage this as two commits, 1 to github/haskoin/haskoin and one to github/haskoin/haskoin-wallet.  I'm a little confused as to the project organization since the files exist in both places.

That being said, this is pullable to master as is, so I am going to make the pull request and wait for feedback.

If it's too messy I'm happy to restage with multiple commits, clean it up to reduce the amount of overall code to bring in, answer questions etc.  I just want to get this into master ;) 

If accepted this fixes #191, #192, #196 and supercedes pull requests #188, #189, and #193.

### DIFF
--- a/haskoin-core/Network/Haskoin/Crypto.hs
+++ b/haskoin-core/Network/Haskoin/Crypto.hs
@@ -2,7 +2,7 @@
   This package provides the elliptic curve cryptography required for creating
   and validating bitcoin transactions. It also provides SHA-256 and RIPEMD-160
   hashing functions; as well as mnemonic keys from BIP-0039.
--}
+-}  
 module Network.Haskoin.Crypto
 (
   -- *Elliptic Curve Keys
@@ -138,28 +138,41 @@ module Network.Haskoin.Crypto
 , deriveMSAddr
 , deriveMSAddrs
 
-  -- ** Custom path derivations
-, DerivPathI((:|), (:/), Deriv, DerivPrv, DerivPub)
-, DerivPath
-, HardPath
-, SoftPath
-, pathToStr
-, parsePath
-, parseHard
-, parseSoft
-, toHard
-, toSoft
-, toMixed
-, (++/), (++|)
-, derivePath
-, derivePubPath
-, derivePathE
+-- ** Custom path derivations
+-- , DerivPathI((:|), (:/), Deriv, DerivPrv, DerivPub)
+, XKeyChildIndex
+, XKeySoftIndex (..)
+, XKeyHardIndex (..)
+, DerivPath (..)
+, HardPath (..)
+, SoftPath (..)
+, Bip32Path (..)
+, Bip32XKey (..)
+, bigWordParser 
+--, pathToStr
+-- , parsePath
+-- , parseHard
+-- , parseSoft
+--, toHard
+--, toSoft
+--, toMixed
+--, (++/), (++|)
+-- , derivePath
+, deriveHardPrvPath
+, deriveSoftPrvPath
+, derivePubPath 
+, derivePathE -- rename deriveBip32Path
+, derivePath 
 
   -- * Custom path address derivations
 , derivePathAddr
 , derivePathAddrs
 , derivePathMSAddr
 , derivePathMSAddrs
+
+, hardPlusSoft
+, softPlusSoft
+, incrementHardPathEnd
 
 ) where
 
@@ -169,4 +182,4 @@ import Network.Haskoin.Crypto.Hash
 import Network.Haskoin.Crypto.Base58
 import Network.Haskoin.Crypto.Mnemonic
 import Network.Haskoin.Crypto.ExtendedKeys
-
+import Network.Haskoin.Crypto.BigWord
--- a/haskoin-core/Network/Haskoin/Crypto/BigWord.hs
+++ b/haskoin-core/Network/Haskoin/Crypto/BigWord.hs
@@ -1,0 +1,174 @@
+module Network.Haskoin.Crypto.BigWord
+(
+-- Useful type aliases
+
+-- Data types
+  BigWord(..)
+, BigWordMod(..)
+
+, Mod31
+, Word31
+
+-- Functions
+, bigWordParser
+) where
+
+{-
+import Test.QuickCheck
+    ( Arbitrary
+    , arbitrary
+    , arbitrarySizedBoundedIntegral
+    )
+-}
+-- import Control.Monad (unless, guard, mzero, (<=<))
+import Control.DeepSeq (NFData, rnf)
+import Data.Bits (Bits(..), FiniteBits(..))
+
+{- 
+
+import Data.Binary (Binary, get, put)
+import Data.Binary.Get
+    ( getWord64be
+    , getWord32be
+    , getWord8
+    , getByteString
+    , Get
+    )
+import Data.Binary.Put
+    ( putWord64be
+    , putWord32be
+    , putWord8
+    , putByteString
+    )
+import Data.Aeson
+    ( Value (String)
+    , FromJSON
+    , ToJSON
+    , parseJSON
+    , toJSON
+    , withText
+    )
+-}
+-- import Data.Ratio (numerator, denominator)
+-- import qualified Data.ByteString as BS (head, length, reverse)
+-- import qualified Data.Text as T (pack, unpack)
+
+-- import Network.Haskoin.Crypto.Curve
+-- import Network.Haskoin.Crypto.NumberTheory
+-- import Control.Monad ((<=<))
+import Network.Haskoin.Util
+import Safe (toEnumMay)
+import Data.String
+import Text.Parsec 
+import Text.Parsec.Language 
+import Text.Parsec.Token
+
+
+type Word31 = BigWord Mod31
+data Mod31
+
+newtype BigWord n = BigWord { getBigWordInteger :: Integer }
+    deriving (Eq, Ord, Read, Show)
+
+instance NFData (BigWord n) where
+    rnf (BigWord n) = rnf n
+
+class BigWordMod a where
+    rFromInteger :: Integer -> BigWord a
+    rBitSize     :: BigWord a -> Int
+
+instance BigWordMod Mod31 where
+    rFromInteger i = BigWord $ i `mod` 2 ^ (31 :: Int)
+    rBitSize     _ = 31
+
+instance BigWordMod n => Num (BigWord n) where
+    fromInteger = rFromInteger
+    (BigWord i1) + (BigWord i2) = fromInteger $ i1 + i2
+    (BigWord i1) * (BigWord i2) = fromInteger $ i1 * i2
+    negate (BigWord i) = fromInteger $ negate i
+    abs r = r
+    signum (BigWord i) = fromInteger $ signum i
+
+instance BigWordMod n => Bits (BigWord n) where
+    (BigWord i1) .&. (BigWord i2) = fromInteger $ i1 .&. i2
+    (BigWord i1) .|. (BigWord i2) = fromInteger $ i1 .|. i2
+    (BigWord i1) `xor` (BigWord i2) = fromInteger $ i1 `xor` i2
+    complement (BigWord i) = fromInteger $ complement i
+    bitSizeMaybe = Just . rBitSize
+    bitSize = rBitSize
+    shift (BigWord i) j = fromInteger $ shift i j
+    testBit (BigWord i) = testBit i
+    bit n = fromInteger $ bit n
+    popCount (BigWord i) = popCount i
+    isSigned _ = False
+    rotate x i = shift x i' .|. shift x (i' - rBitSize x)
+      where
+        i' = i `mod` rBitSize x
+
+instance BigWordMod n => FiniteBits (BigWord n) where
+    finiteBitSize = rBitSize
+
+instance BigWordMod n => Bounded (BigWord n) where
+    minBound = fromInteger 0
+    maxBound = fromInteger (-1)
+
+instance BigWordMod n => Real (BigWord n) where
+    toRational (BigWord i) = toRational i
+
+instance BigWordMod n => Enum (BigWord n) where
+    succ r@(BigWord i)
+        | r == maxBound = error "BigWord: tried to take succ of maxBound"
+        | otherwise = fromInteger $ succ i
+    pred r@(BigWord i)
+        | r == minBound = error "BigWord: tried to take pred of minBound"
+        | otherwise = fromInteger $ pred i
+    toEnum i
+        | toInteger i >= (toInteger minW) && toInteger i <= (toInteger maxW) = r
+        | otherwise = error "BigWord: toEnum is outside of bounds"
+      where
+        f i = if toInteger i >= (toInteger minW) && toInteger i <= (toInteger maxW)
+                then Just r
+                else Nothing
+        minW = minFrom r
+        maxW = maxFrom r
+        r = fromInteger . toEnum $ i
+        minFrom :: BigWordMod a => BigWord a -> BigWord a
+        minFrom _ = minBound
+        maxFrom :: BigWordMod a => BigWord a -> BigWord a
+        maxFrom _ = maxBound
+    fromEnum (BigWord i) = fromEnum i
+
+
+instance BigWordMod n => Integral (BigWord n) where
+    (BigWord i1) `quot` (BigWord i2) = fromInteger $ i1 `quot` i2
+    (BigWord i1) `rem` (BigWord i2) = fromInteger $ i1 `rem` i2
+    (BigWord i1) `div` (BigWord i2) = fromInteger $ i1 `div` i2
+    (BigWord i1) `mod` (BigWord i2) = fromInteger $ i1 `mod` i2
+    (BigWord i1) `quotRem` (BigWord i2) = (fromInteger a, fromInteger b)
+      where
+        (a,b) = i1 `quotRem` i2
+    (BigWord i1) `divMod` (BigWord i2) = (fromInteger a, fromInteger b)
+      where
+        (a,b) = i1 `divMod` i2
+    toInteger (BigWord i) = i
+
+
+-- TODO: Test
+
+instance ToHaskoinString (BigWord n) where
+    toHaskoinString (BigWord i) = show i
+instance (BigWordMod n) => FromHaskoinString (BigWord n) where
+    fromHaskoinString = eitherToMaybe . parseBigWord -- toEnumMay <=< readMay
+instance (BigWordMod n) => IsString (BigWord n) where
+    fromString = derivFromString "Could not decode extended private key: " 
+
+parseBigWord :: (BigWordMod n) => String -> Either ParseError (BigWord n)
+parseBigWord x = parse bigWordParser "bigWordParser" x
+bigWordParser :: (BigWordMod n) => Parsec String () (BigWord n)  
+bigWordParser = do
+  i <- integer . makeTokenParser $ haskellStyle 
+  case toEnumMay . fromIntegral $ i of
+    Just w -> return w
+    Nothing -> fail $ "parseBigWord, " ++ show i
+
+
--- a/haskoin-core/Network/Haskoin/Crypto/ExtendedKeys.hs
+++ b/haskoin-core/Network/Haskoin/Crypto/ExtendedKeys.hs
@@ -77,7 +77,6 @@ import Data.Binary.Get (Get, getWord8, getWord32be)
 import Data.Binary.Put (Put, putWord8, putWord32be)
 import Data.Word (Word8, Word32)
 import Data.Bits (setBit, testBit, clearBit)
--- import Data.List.Split (splitOn)
 import Data.Maybe (fromMaybe)
 import Data.String (IsString, fromString)
 import Data.String.Conversions (cs)
@@ -92,8 +91,6 @@ import Network.Haskoin.Script.Parser
 import Network.Haskoin.Crypto.Keys
 import Network.Haskoin.Crypto.Hash
 import Network.Haskoin.Crypto.Base58
-import Data.List (foldl')
--- import Network.Haskoin.Crypto.ExtendedKeyHelpers
 import Network.Haskoin.Crypto.BigWord (Word31, bigWordParser    )
 import Data.Monoid (Endo(..), (<>))
 import Text.Parsec
@@ -477,7 +474,9 @@ data HardPath = (:/|) !XKeySoftIndex !HardPath
                 | (:||) !XKeyHardIndex !HardPath
   deriving (Read,Show,Eq,Ord)
 
+incrementSoftIndex :: XKeySoftIndex -> XKeySoftIndex
 incrementSoftIndex (XKeySoftIndex i) = XKeySoftIndex $ i + 1
+incrementHardIndex :: XKeyHardIndex -> XKeyHardIndex
 incrementHardIndex (XKeyHardIndex i) = XKeyHardIndex $ i + 1
 
 incrementHardPathEnd :: HardPath -> HardPath
@@ -489,13 +488,8 @@ incrementHardPathEnd ( (:|/) hardIndex nonEmptySoftPath ) = hardIndex           
 
 incrementSoftPathEnd :: SoftPath -> SoftPath -> SoftPath
 incrementSoftPathEnd def XKeyEmptyPath            = def
-incrementSoftPathEnd def ( (://) softIndex XKeyEmptyPath) = incrementSoftIndex softIndex :// XKeyEmptyPath
-incrementSoftPathEnd def ( (://) softIndex softPath)             = softIndex :// incrementSoftPathEnd def softPath
-
-modifySoftEndOfSoftPath :: (XKeySoftIndex -> XKeySoftIndex) -> SoftPath -> Either String SoftPath
-modifySoftEndOfSoftPath f XKeyEmptyPath = Left "modifySoftEndOfSoftPath, empty soft path" 
-modifySoftEndOfSoftPath f ( (://) i XKeyEmptyPath ) = Right $ (f i) :// XKeyEmptyPath
-modifySoftEndOfSoftPath f ( (://) i p ) = (i ://) <$> modifySoftEndOfSoftPath f p
+incrementSoftPathEnd _ ( (://) softIndex XKeyEmptyPath) = incrementSoftIndex softIndex :// XKeyEmptyPath
+incrementSoftPathEnd def ( (://) softIndex softPath)    = softIndex :// incrementSoftPathEnd def softPath
 
 instance NFData HardPath where
     rnf ( (:/|) index path ) = rnf index `seq` rnf path
@@ -503,6 +497,7 @@ instance NFData HardPath where
     rnf ( (:||) index path ) = rnf index `seq` rnf path
     
 instance NFData SoftPath where
+    rnf XKeyEmptyPath = () 
     rnf ( (://) index path ) = rnf index `seq` rnf path
 
 instance NFData DerivPath where
@@ -628,17 +623,17 @@ hardPathParser = do
   where hardThenSoft, softThenHard, hardThenHard :: Parsec String () HardPath
         hardThenSoft = do
             i <- hardIndexParser            
-            string "/"
+            _ <- string "/"
             p <- softPathParser
             return $ i :|/ p
         softThenHard = do
             i <- softIndexParser
-            string "/"
+            _ <- string "/"
             p <- hardPathParser
             return $ i :/| p
         hardThenHard = do
             i <- hardIndexParser
-            string "/"
+            _ <- string "/"
             p <- hardPathParser
             return $ i :|| p
         oneHardIndex = do
@@ -657,7 +652,7 @@ parseHardIndex x = parse hardIndexParser "hardIndexParser" x
 hardIndexParser :: Parsec String () XKeyHardIndex
 hardIndexParser = do
     i <- bigWordParser
-    string "'"
+    _ <- string "'"
     return $ XKeyHardIndex i
 
 
@@ -689,12 +684,12 @@ parseSoft x = parse softPathParser "softPathParser" x
 softPathParser :: Parsec String () SoftPath
 softPathParser = do
   softPath <- try more <|> try oneSoftIndex <|> try ( return XKeyEmptyPath )    
-  eof <|> ( try $ string "/" >> eof ) -- optional trailing slash
+  eof <|> ( try $ do _ <- string "/" ; eof ) -- optional trailing slash
   return softPath
   where more :: Parsec String () SoftPath
         more = do 
             i <- softIndexParser
-            string "/"
+            _ <- string "/"
             p <- softPathParser
             return $ i :// p
         oneSoftIndex = do
@@ -779,9 +774,9 @@ deriveBip32Path dp xkey =
     Bip32PubK xpbk -> 
       case dp of
         Bip32PubM (Bip32Soft path) -> Right . Bip32PubK . dXPPubPath path $ xpbk
-        Bip32PubM (Bip32Hard path) -> Left $ "deriveBip32Path, can't derive hard path starting with public key for path: " ++ show dp ++ ", xkey: " ++ show xkey
-        Bip32Prvm (Bip32Soft path) -> Left $ "deriveBip32Path, can't derive private key starting with public key for path: " ++ show dp ++ ", xkey: " ++ show xkey
-        Bip32Prvm (Bip32Hard path) -> Left $ "deriveBip32Path, can't derive private key starting with public key for path: " ++ show dp ++ ", xkey: " ++ show xkey
+        Bip32PubM (Bip32Hard path) -> Left $ "deriveBip32Path, can't derive hard path starting with public key for path: " ++ toHaskoinString path ++ ", xkey: " ++ toHaskoinString xkey
+        Bip32Prvm (Bip32Soft path) -> Left $ "deriveBip32Path, can't derive private key starting with public key for path: " ++ toHaskoinString path ++ ", xkey: " ++ toHaskoinString xkey
+        Bip32Prvm (Bip32Hard path) -> Left $ "deriveBip32Path, can't derive private key starting with public key for path: " ++ toHaskoinString path ++ ", xkey: " ++ toHaskoinString xkey
     Bip32PrvK xpvk -> 
       case dp of
         Bip32PubM (Bip32Soft path) -> Right . Bip32PubK . deriveXPubKey . dXPrivPath path $ xpvk
--- a/haskoin-core/Network/Haskoin/Crypto/ExtendedKeys.hs
+++ b/haskoin-core/Network/Haskoin/Crypto/ExtendedKeys.hs
@@ -571,16 +571,13 @@ parsePath :: String -> Either ParseError DerivPath
 parsePath x = parseConsumingAll derivPathParser "derivPathParser" x
 
 derivPathParser :: Parsec String () DerivPath
-derivPathParser = do
-    dp <- try ( Bip32Prvm <$> 
-                 ( ( try $ ( string "m/" >> bip32PathParser )) 
-                   <|> ( (try $ string "m") >> return (Bip32Soft XKeyEmptyPath) ) )
-              ) 
-      <|> try ( Bip32PubM <$> 
-                ( ( try $ string "M/" >> bip32PathParser) 
-                  <|> ( (try $ string "M") >> return (Bip32Soft XKeyEmptyPath) ) )
-              )
-    return dp
+derivPathParser = 
+    choice [  Bip32Prvm <$> choice [ try $ string "m/" >> bip32PathParser,  
+                                     try $ string "m"  >> return empty ], 
+              Bip32PubM <$> choice [ try $ string "M/" >> bip32PathParser, 
+                                     try $ string "M"  >> return empty ]
+            ]
+  where empty = Bip32Soft XKeyEmptyPath
 
 instance ToHaskoinString Bip32XKey where
     toHaskoinString x = case x of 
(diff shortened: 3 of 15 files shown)


tphyahoo:
I peeked at travis results and noticed there are some unhandled warnings in the compile.

I suppose these should be squashed.

m0dev76:
Hi tphyahoo and thanks for the pull request. Here are some of my comments on the code so far.

In my original code, I am using a phantom type `DerivPathI t` where `t` can be one of `soft`, `mixed`, or `hard`. `mixed` is the generic type whereas `soft` and `hard` are more type strict. The goal of using phantom types here is to leverage the type system and give the compiler more meaningful information about the purpose of our program at compile time. For example, I can write:

```
λ> DerivPrv :/ 1 :/ 2 :/ 3 :: SoftPath
SoftPath "m/1/2/3"
```

for a derivation path that contains only soft derivations. I am telling the compiler that my code here only contains soft paths. This truth doesn't need to be determined at runtime. You can't fool the compiler:

```
λ> DerivPrv :/ 1 :| 2 :/ 3 :: SoftPath
<interactive>:8:15:
    No instance for (HardOrMixed
                       Network.Haskoin.Crypto.ExtendedKeys.Soft)
      arising from a use of ‘:|’
    In the first argument of ‘(:/)’, namely ‘DerivPrv :/ 1 :| 2’
    In the expression: DerivPrv :/ 1 :| 2 :/ 3 :: SoftPath
    In an equation for ‘it’: it = DerivPrv :/ 1 :| 2 :/ 3 :: SoftPath
```

If I want to have a mix of soft and hard paths, I need to explain that to the compiler through the appropriate type `DerivPathI Mixed` which also has a type alias for `DerivPath`.

```
λ> DerivPrv :/ 1 :| 2 :/ 3 :: DerivPath
DerivPath "m/1/2'/3"
```

I do believe that exposing most of the logic to the type system is a good thing, and phantom types are an interesting way to leverage the type system. Down the road, when I'm writing user apps and wallets for example, and I specify a database type as `HardPath`, I know that I can only have hard paths in there and I will never be able to compile some code that tries to squish something else into my database. My code will never crash at runtime because I'm using a soft derivation instead of a hard one. It may crash due to other reasons but not that one :)

When you define `HardPath` as you did here:

```
data HardPath = (:/|) !XKeySoftIndex !HardPath
                | (:|/) !XKeyHardIndex !SoftPath
                | (:||) !XKeyHardIndex !HardPath
```

then you can effectively build something of type `HardPath` which contains soft derivations and thus, I think, we are loosing type safety. As a side note, I think it's much easier for library users to expose only the `(:/)` and `(:|)` operators as I have in the original code.

With respect to allowing hard-follow-soft derivations, please see my pull request #201 for an example of what I had in mind. The pull request is probably not complete but its simply to offer as a counter example to your current approach.

tphyahoo:
> When you define HardPath as you did here... then you can effectively build something of type HardPath which contains soft derivations

I don't see how.  Can you give an example?

m0dev76:
```
λ> (XKeyHardIndex 3) :|/ ((XKeySoftIndex 2) :// ((XKeySoftIndex 1) :// XKeyEmptyPath)) :: HardPath

(:|/) (XKeyHardIndex (BigWord {getBigWordInteger = 3})) ((://) (XKeySoftIndex (BigWord {getBigWordInteger = 2})) ((://) (XKeySoftIndex (BigWord {getBigWordInteger = 1})) XKeyEmptyPath))
```

Additionally, you have to specify the derivations in reverse. The above code is to specify a 1/2/3' derivation.

tphyahoo:
But this is a hard path.  What's the problem?

re Derivation in reverse -- maybe you have a point there.  I didn't intend to have things reversed if so.  Oversight. I think easy fix.

tphyahoo:
My understanding: the invariant you want to enforce is that you can't derive a pub key from a pubkey, given a hard path.

Anything that satisfies this is a legitimate hard path.  So the above hard path is legitimate.

tphyahoo:
Re the direction flipping: I have to go do weekendy stuff, but I will revise

https://github.com/tphyahoo/haskoin/blob/bip32_paths_mergecandidate1/haskoin-core/tests/Network/Haskoin/Crypto/ExtendedKeys/Units.hs

and relevant code, with things reading the correct direction.

One thing to bear in mind is that bip32 paths are read left to right, whereas usual function application is right to left.

Maybe this is a situation where Control.Category and >>> instead of <<< could make things more legible.

Cheers.

tphyahoo:
I will also note that my update fails too-large indexes at compile time, by using 31 bit integer type, if the indexes are hard coded.

m0dev76:
OK, I think we've been working with different definitions for hard and soft path. I've defined `HardPath` as a path containing only hard derivations and `SoftPath` as a path containing only soft derivation.

The rationale behind choosing those definitions came from writing the wallet. In the wallet database, we store the derivation path of accounts and the derivation path of addresses (relative to an account). Ideally, we want account to have only hard derivation in their path. We could argue whether soft derivations in the middle are acceptable or not, but simply to be on the safe side, we want only hard derivations. Thus I defined the type `HardPath` to mean "only hard derivations". 

Then, when dealing with derivation paths for addresses (relative to an account), I want there to be only soft derivations, otherwise I will not be able to derive the addresses from the account XPubKey. Thus I defined the `SoftPath` type to mean "only soft derivations". So, with these definitions, the only way to derive a pubkey from a pubkey is to use a `SoftPath` derivation. Anything that contains both hard derivations and soft derivations is a mixed derivation path.

Now if we look at our definitions closely, I notice that they are essentially the same, except that your definition of `HardPath` encompasses both my definition of `HardPath` and mixed path (`DerivPathI Mixed`). We could debate as to which definition is the best and I'm sure both point of views can be defended. But I would tend to err towards using different types for `HardPath` and mixed path unless you don't find my use case for account derivations (only hard) above valid.

m0dev76:
If we wanted to squash hard and mixed path into one type, I would propose to use a definition similar to this:

```
data Hard
data Soft

data DerivPathI t where
    (:|) :: DerivPathI t -> KeyIndex -> DerivPathI Hard
    (:/) :: DerivPathI t -> KeyIndex -> DerivPathI t
```

That way, if I ever use `(:|)` in a derivation, it will automatically become a `HardPath`. The only way to get a `SoftPath` would be to have only soft derivations `(:/)`. So:

```
λ> DerivPrv :/ 1 :/ 2 :/ 3 :: SoftPath
SoftPath "m/1/2/3"
```

would work but

```
λ> DerivPrv :/ 1 :| 2 :/ 3 :: SoftPath
error ...
```

would fail. However, a mixed path would now be a valid `HardPath` type:

```
λ> DerivPrv :/ 1 :| 2 :/ 3 :: HardPath
HardPath "m/1/2'/3"
```

That's one possibility I am willing to consider. I guess it all depends on whether having a separate type in the type system for "only hard derivations" is useful.

tphyahoo:
Thanks for that, it getting a better idea of the landscape.

We have a terminology conflict.  Since yours came first, I will try to say HardPath to refer to the current HardPath in master, and call my new stuff Bip32HardPath, Bip32SoftPath etc. If I proceed on this path I will probably rename the types as well.  You're right, they refer to different concepts.  Mine is more concerned with enforcing the Bip32 spec for a general purpose library, current is more about servicing the specific needs of haskoin-wallet.

My objectives for Bip32Hard / Bip32Soft
-- follow the bip32 spec (which has 3/2'/1 type mixed paths).
-- enforce the invariants of the bip32 spec in the compiler.

Enforcing invariants is key.

Specifically, I don't think there are any invariants enforced by HardOrMixed, or Mixed in general, and that's Bad. 

Or what would you say is the purpose of Mixed in the current setting?

tphyahoo:
Another thing.

I am still familiarizing myself with haskoin-wallet so I'm not certain but..

> Then, when dealing with derivation paths for addresses (relative to an account), I want there to be only soft derivations, otherwise I will not be able to derive the addresses from the account XPubKey.

For addresses, is it not the case that the soft derivation is always a single soft path segment?

Just to lobby a bit, note that in my scheme Bip32SoftPath and Bip32SoftIndex (renamed per previous comment) both implement XPubKeyEndoDerivable.  And do _not_ implement XPrvKeyEndoDerivable, which enforces the invariant, in case that was unclear.

So if addresses always map to a single path segment (which is to say just an index) you could use XKeySoftIndex if that makes the intent of the code more clear and explicit.  

Just an example to show how using classes to enforce the invariants gives you additional flexibility.

tphyahoo:
>  I guess it all depends on whether having a separate type in the type system for "only hard derivations" is useful.

It is useful, because it enables you to enforce an important invariant, with hard security implications, at compile time.

tphyahoo:
So, to recap, we are interested in enforcing slightly different invariants.

HardPath -- only hard segments.  Don't care about supporting mixed paths.
Bip32HardPath -- allow mixed paths.  Don't allow pub -> pub derivations. 
HardOrMixed -- ?? 
MixedOrSoft -- ??

I think the mixed cases don't enforce anything, which makes it (I think unnecessarily) harder to follow the logic of the code.

A possible way forward:

I work to get Bip32HardPath / Bip32SoftPath integrated, and allowing all Bip32 spec cases, including ones not currently supported.

HardPath / SoftPath remain, for enforcing the invariants wanted for Haskoin Wallet (all hard or soft paths). If you stick all the actual pub->pub, prv->prv derivation stuff under new Bip32HardPath / Bip32SoftPath machinery, the code for HardPath / SoftPath gets much simplified.  (No more phantom type need.)

We drop HardOrMixed, MixedOrSoft, because this doesn't really provide anything.

Side notes:

Bip32SoftPath is Monoid with empty soft path as zero, and Bip32HardPath is a SemiGroupoid.  addition is obvious append operation.  Bip32HardPath doesn't have an identity because hard paths need at least one element. (I didn't actually provide these instances, maybe I should.)

It's much harder to wrangle the 

> data DerivPathI t

stuff into monoid / semigroupoid, because you mix the path stuff with the m/M tagging at the beginning of the path.

So that's another point of elegance for the alternative implementation, which decouples the m/M markers.

With the original code, I am still thinking about what to do with Deriv, DerivPrv and DerivPub.  

I don't understand what Deriv buys you.

m0dev76:
If you consider the code from #201, the only way to derive keys from a derivation path is by using either `derivePath`, `derivePubPath` or `derivePathE`. They have the following type definitions:

```
derivePath :: DerivPathI t -> XPrvKey -> XPrvKey
derivePubPath :: SoftPath -> XPubKey -> XPubKey
derivePathE :: DerivPathI t -> XPrvKey -> Either XPubKey XPrvKey
```

`derivePath` is the general version from which you can derive a private key given a private key and any derivation path (hard, mixed soft). Arguably, that is correct and supported by the BIP32 spec.

`derivePubPath` is the only function you can use to derive a public key from a public key. And you can only use a `SoftPath` as a derivation path here. Trying to match any other types will obviously create an error at compile time. I think this function provides the same invariant that you are describing in your previous post. 

`derivePathE` is a helper function if you are using the "m/" or "M/" notations for specifying your derivations. In such a case, you don't know at compile-time if you want a private or public key. You only know at runtime after parsing the given string. But this functions only takes private keys as an input so you will never have a situation where you will try to derive a public key from a public key. Maybe different type signatures can be explored for this function. I'm open to suggestions.

Now all this derivation path business is a higher level concept that reuses 3 functions that perform the basic derivations. Arguably, a library user can use these lower level functions only. They are:

```
prvSubKey :: XPrvKey -> KeyIndex -> XPrvKey
pubSubKey :: XPubKey -> KeyIndex -> XPubKey
hardSubKey :: XPrvKey -> KeyIndex -> XPrvKey
```

These 3 functions rigorously implement the bip32 specifications and their types do not allow for much ambiguity. But you can only do one derivation at a time using them so the concept of defining derivation paths is a useful one.

So in fairness, I do believe that the code follows the bip32 specifications. It didn't allow for hard-follow-soft derivations but that is only a small fix to tweak (see #201). If you have a `HardPath`, you can derive private or public keys from private keys using `derivePath` or `derivePathE` but you can not derive public keys from public keys through `derivePubPath`. 

If you have a mixed path (`DerivePathI Mixed`), it is basically a relaxed type (it could be mixed or hard or soft but the compiler doesn't know so it labels it as mixed). You can't derive a public key from a public key using a mixed derivation. It doesn't type check with `derivePubPath`. You can only use a mixed path in the same way you can use a `HardPath` by deriving things with private keys. The only difference between a mixed path and a `HardPath` is that a `HardPath` is a type that guarantees that only hard derivations are present in the data. 

Unless there are other invariants that I am unaware of, I think the code covers them. The typeclasses `HardOrMixed` and `MixedOrSoft` are not there to enforce the invariants but are there to constraint the valid types that `t` can have. The invariants are enforced by the functions that I described above. The typeclasses are nice tricks to allow a simple syntax for the derivations. If I use `(:|)`, then the phantom type for `t` can be either `Hard` or `Mixed` which makes sense. And if I use `(:/)` then the phantom type for `t` can be either `Mixed` or `Soft` which also makes sense. Now the magic happens when you combine them together. The type classes will restrict the valid values that `t` can have. For example, what are the valid values for `t` if this example?:

```
Deriv :| 1 :| 2 :| 3
```

It's important to note that `t` for all 3 derivations has to be the same. So the first derivation enforces either `t` is `Hard` or `t` is `Mixed`. The same applies for the second and third derivation. So you have two valid types for this expression:

```
Deriv :| 1 :| 2 :| 3 :: HardPath
Deriv :| 1 :| 2 :| 3 :: DerivPath
```

(`DerivPath` is a type alias for `DerivPathI Mixed`). Now this looks fine because it is indeed a valid `HardPath` and also a valid mixed path. Remember that mixed actually means _unknown_, so you're not telling the compiler what really is there. 

Now what happens if you mix stuff up? Consider:

```
Deriv :| 1 :/ 2 :| 3
```

What are the valid types for `t` here? `t` is constrained to `Hard` or `Mixed` by the first derivation and third derivation. The second derivation constrains `t` to `Mixed` or `Soft`. So what are the valid possibilities  for `t`? The only valid value for `t` is `Mixed`. If you try to stuff that expression into a `HardPath` or `SoftPath` it will fail. That is also arguably correct because this expression is neither a `HardPath` or a `SoftPath`. 

This is the ultimate purpose of those two typeclasses `HardOrMixed` and `MixedOrSoft`. They serve to constrain the valid values for the phantom type `t` when building derivation paths. They don't directly enforce the bip32 invariants, those are enforced through the functions.

m0dev76:
`Deriv` allows to store a generic derivation path which is neither public or private. It could be both. This is useful for example to store the derivation path of an address in the wallet. Sometimes you need to generate the address itself, so that is a public derivation. Sometimes you need to derive the private key (for example when spending some coins), so that is a private derivation. The way I write that down is simply "/1/2/3" instead of "m/1/2/3" or "M/1/2/3".

m0dev76:
As a side note, `HardOrMixed` and `MixedOrSoft` are not intended to be exposed to library users. They are simply internal helpers that make sure that the types of the derivations are always coherent. If the user tries to do invalid things like `Deriv :/ 3 :| 4 :: HardPath`, the typeclass constraints will be broken and the code will crash at compile time. But the user doesn't need to know those typeclasses are there.

m0dev76:
In a sense, I'm using the typeclasses as constraints of the phantom type `t` to create a shorter syntax for derivations. I only have 2 operators `(:|)` and `(:/)` instead of 4 `(:||)`, `(:|/)`, `(:/|)`, `(://)`. Without the typeclasses on the phantom types, you are forced to spell out explicitly all the combinations of hard/soft paths which yields 4 data constructors.

tphyahoo:
Another thing.

I believe the current system conflates hardness vs softness with private versus public.

That's because the Deriv values begin with

> Deriv   or DerivPrv  or DerivPub  

I'm not certain, but I believe derivations beginning with DerivPub are constrained to be Soft.

Is that the case? If so, might be this is the root of the problem in #196.

By decoupling hard vs soft from private vs public, I completely avoid this confusion.

m0dev76:
There's nothing preventing you from writing `DerivPub :| 1 :| 2` or `DerivPub :| 1 :/ 2`. The data constructors `DerivPrv` and `DerivPub` are only there to encode the information that we have seen a "m/" or a "M/" while parsing a String. That's pretty much all. It is only used for nice display of the string, and by `derivPathE` to use that information to produce the right final key (public or private).

m0dev76:
It is very uncommon that you want to store, for example in a database, the full string `m/1/2'/3` because ultimately that notation represents a function application. `m/` suggests that you want to apply a private key to a derivation path to get another private key. However, most of the time, we are really only interested in storing the path itself. How we deal with the path (if we want a `m/` or a `M/`) is defined by which functions we call on our derivation (`derivPath` or `derivPubPath`). Its very strange to store a function application in a database if you think about it. You typically store data and apply your data to functions. The current solution was designed with this in mind, which is why the generic `Deriv` constructor was provided, to be able to simply express an arbitrary derivation path (the data part) without having to specify the function application part (if we want a public or a private key from our derivation path).

However, if you are really only interested in parsing a derivation string and use it directly as a function, you can use: `parsePath` which will produce a derivation path containing the correct constructor `DerivPrv` or `DerivPub`. Then you can call `derivPathE` with a private key to "apply" the function which was implicit in the string you just parsed to the private key. But I don't do this very often in practice. I usually don't care about the `m/` or the `M/` in a derivation path. I'm typically more interested in simply storing the derivation data itself and decide later if I want to have a `m/` or a `M/` through simple pure functions in haskell.

Putting it another way, a notation like `m/1/2/3` doesn't represent data, it represents a function. You would have to be quite acrobatic to try and represent a function application as data types and type classes in haskell. This is why I took a simpler approach to the problem. I split the concept of `m/1/2/3` into 2 parts: `m/` which is a function and `1/2/3` which is data. Then it's very simple to map to haskell: `m/` becomes a pure function and `/1/2/3` becomes a data type. That is the rationale behind the current solution.

m0dev76:
Regarding issue #196, it would be easy to fix by replacing `derivPath` in the test by `derivPathE`. The test is working as intended in the sense that `derivPath` is basically slapping `m/` in front of your derivation. As I explained in the previous post, the solution has been mapped in haskell such that `m/` and `M/` are pure functions in haskell, so `m/` is basically `derivPath` and `M/` is `derivPubPath`. You can use `derivPathE` if you want to use the information contained in the string directly. The test should probably use `derivPathE` and the expected result would be achieved. 

I don't think that a solution to this problem where we try to encode the full function application together with the derivation data into data types and type classes will yield a good result. Things which are naturally functions (like `m/` and `M/`) should be mapped as functions I believe, and not warped into data unless there is really no other solutions.

jprupp:
Hello,

I have been following the whole conversation, and I will side with @plaprade on this one. @plaprade's solution, which is to keep the current code and relax the restrictions on hard-follows-soft solves aligns the code fully with the BIP-32 specification, does not introduce extra complexity (endomorphisms and corresponding typeclasses), and preserves type-safety.

Type-safety deal is important. Thanks to `SoftPath` and `HardPath`, sensitive code dealing with BIP-32 derivations can be checked at compile-time as @plaprade indicated, with minimal cognitive load, making it safer to write correct HD wallet code. Haskoin Wallet is but an implementation of an HD wallet that benefits from this.

I would also avoid resurrecting `BigWord` just for this case, and use bound checks to verify that the derivation index falls within 31 bits.

I think we should build on #201 until it passes CI checks. This PR is too complex.

m0dev76:
I understand why there could be a bit of confusion regarding the `DerivPrv` and `DerivPub` data constructors. Originally, I didn't intend for them to exist at all. My intention was only to model the derivation path itself without the "m/" or "M/" part so only the /1/2'/3 part. Whether you wanted a private key or a public key would depend on which functions you used with your derivation path.

However, that lead to one slight problem. When you parsed an arbitrary path from a string, you would loose the information on whether you have seen an "m/" or an "M/". I decided to add these two data constructors to record what has been parsed. They are not there to enforce anything however, they are only informational. The `derivePathE` function does use this information to derive the proper key.

I just wanted to point this out as the idea behind this library was really to provide a good representation of the /1/2'/3 part of the derivation and let the user choose if he needs private or public keys through functions. I do believe that it provides a simple api to work with. Throughout development, I have continuously redesigned this api to be as easy to use as possible while I was working on the wallet.

tphyahoo:
OK, I'll concentrate on the test cleanup stuff and the other bip32 issues I already pointed out, and soft pedal this approach.  It is certainly a beast of a PR.

m0dev76:
Additionally, I do not see the use of phantom types as something bad. It is a neat way to work with with different types under the same symbol. Without phantom types, you often end up having to define wrapper types like:

```
data DerivPath = Bip32Prvm Bip32Path | Bip32PubM Bip32Path
data Bip32Path = Bip32Hard HardPath | Bip32Soft SoftPath 
```

which makes it arguably a bit harder to work with the library from a user perspective.

However, as I have been reflecting on the way that I have been encoding `DerivPrv` and `DerivPub`, I do appreciate your solution of using a super type around a path to encode this information. I'm thinking of something like this:

```
type DerivPath = DerivPathI Mixed

data DerivPathI t where
    (:|)     :: HardOrMixed t => !(DerivPathI t) -> !KeyIndex -> DerivPathI t
    (:/)     :: MixedOrSoft t => !(DerivPathI t) -> !KeyIndex -> DerivPathI t
    Deriv :: DerivPathI t

data ParsedPath = ParsedPub DerivPath | ParsedPrv DerivPath
```

You wouldn't use `ParsedPath` to build anything in Haskell directly. It would probably only be used when parsing a raw string to record the "m/" or "M/" bit. But it removes the `DerivPrv` and `DerivPub` from `DerivPathI` which I never really liked.

tphyahoo:
xenog: 

> "extra complexity (endomorphisms and corresponding typeclasses)"

Endomorphism captures the notion of transformation within the same type. (a -> a). So 

> XPubKey -> XPubKey (XPubKeyEndo) or XPrvKey to XPrvKey (XPrvKeyEndo)

The type classes just state which path types allow which endomomorphisms are allowed on which path type.

Bip32SoftPath is both XPubKeyEndo and XPrvKeyEndo.  Bip32HardPath is only XPrvKeyEndo.

I have more trouble with the phantom type stuff.

I also think Monoid / Semigroupoid mentioned above might be useful.

I'm still hoping some of this intuition can be captured somehow, even if it means starting from scratch with #201.  I think the alternate 

> data DerivPathI t

proposed by plaprade above is a promising start point.

m0dev76:
I've updated #201 with what I had in mind for separating `DerivPrv` and `DerivPub`. The code is now well separated in two parts. The first part only deals with representing a derivation path in Haskell and does no parsing whatsoever (only deals with representing /1/2'/3). The `DerivPathI t` type can deal with hard, soft or generic types in the type system. I've renamed `Mixed` to `Generic` as this was a point of confusion. `Generic` conveys much better the information that the real type (hard or soft) is not known to the compiler. 

The second part of the code introduces a `ParsedPath` type which deals with parsing a derivation string and storing the information about wanting a private (m/) or public (M/) key. A single function is provided (`applyPath`) that tries to apply the result of parsing a bip32 string to a key. I think this is similar to what you have proposed in this pull request.

I haven't introduced any other concepts (monoids, endomorphisms) as I want to have a small diff. I'm open to suggestions for providing type class instances if they are useful and keep the user experience simple.

tphyahoo:
#201 is better, but I'm still shaky on some aspects.

First, is there a way I can load this in ghci, since stack build currently fails? (I'm new to stack and the only way I know to get into ghci is stack ghci, but this needs stack build first, to work.)

OK, so I couldn't play with paths in ghci, so forgive any errors in the following please.

My question is, which of the following typechecks?

```
Deriv :: HardPath
Deriv :: SoftPath
```

I believe both type check.  And, I see a problem here.

Deriv amounts to the empty path.

HardPath definitionally is something that you cannot derive the result, starting from a public key and path.

But with the "empty" path, you _can_ derive pubkey from pubkey + path, it's just the same key, and the path (the empty path / Deriv) maps to  identity function.

So I would argue that Deriv should be SoftPath but not HardPath. (Which is the moral equivalent of what I had in current pull request.)

Because Deriv satisfies both hard and soft path, we are blocked from creating classes from privkey/pubkey endos, that satisfy the bip32 invariants. (Namely: Pubkey uncalculatable given only pubkey and HardPath.)

Similarly HardPath is now I think a monoid (with an identity) when really it should be a SemiGroupoid with no identity. (SoftPath is still a monoid with Deriv as identity.)

There are actual security consequences to this.  Programmer could have code that typechecks as HardPath, but it's just the empty path, and they don't realize the public key can be derived from the path in this instance.

Of course, you can just say "who cares, this is a tiny little edge case." 

I'm just bringing this up in case this argument changes anything in your thinking.

Cheers.

m0dev76:
Sorry about the build. I've fixed the tests for #201 so they build. Some of the tests don't pass however but I'll look into it when we've nailed down the style.

I understand your issue concerning `Deriv`. It is true that `Deriv` currently satisfies all of those three properties:

```
Deriv :: HardPath
Deriv :: DerivPath
Deriv :: SoftPath
```

The way I look at it is that an empty path is essentially an identity function and it doesn't matter much which type it has. I don't think that, from a conceptual point of view, it breaks anything. If you decide to have `Deriv :: HardPath` then you can't call `derivePubPath :: SoftPath -> XPubKey -> XPubKey` because of the type mismatch. If you define `Deriv :: SoftPath` then you can call `derivePubPath` but it's essentially just going to act as `id` so its really not very interesting. Similarily, using `Deriv` with `derivePath` is simply going to yield the `id` function. 

I'm not entirely sure I follow your argument about the security issue (its very late here so maybe my brain is a bit fried). If a user has some field in a database for example that is of type `HardPath`, then there will only be derivations there that are all hard or the empty derivation `Deriv` which can also be of type `HardPath`. In both cases, you can't call `derivePubPath` so you're really not at risk of doing anything with a public key. You can't cast your `Deriv :: HardPath` to a `SoftPath` so that it can get squished into `derivePubPath` unless you explicitly pass it through a casting function like `toSoft`. But if the developer does that, he likely knows what he is doing. And `toSoft` will return `Nothing` anyway if the given path is not truly a `SoftPath`. And, ultimately, the worst thing that can happen is that a programmer manages to cast his empty derivation `Deriv :: HardPath` into `Deriv :: SoftPath` and then injects it into `derivePubPath` which will then only yield the `id` function so from a security perspective I don't think that is really a risk. You're just going to get back the public key that you've injected. 

Now from a purity perspective, if there is an issue with having `Deriv` satisfy multiple types (like the inability to define a prvkey/pubkey endomorphism), then maybe we can have a look at some alternatives. We could define different empty paths like `DerivSoft` or `DerivHard` but I wonder if that's not going to be confusing for users. If we make `Deriv` only of type `Deriv :: SoftPath` then I'm afraid that we won't be able to use the flexibility of the phantom type system to derive the correct type automatically depending on the users choice of constructors `(:|)` or `(:/)`. `Deriv :| 0` would be invalid for example because `(:|)` imposes a `HardOrGeneric` constraint on the type but we force it to `SoftPath` so its impossible to resolve a matching type. We could impose a `SoftOrGeneric` constraint on `Deriv` but then all types will always have to be `Generic` or `Soft` and we would never be able to have `Hard`.

Let me know if you have a different solution in mind. I think that we have a situation here where we kind of have to find the right compromise between usability and mathematical elegance.

m0dev76:
To break it down from a high level point of view, your approach is trying to capture the bip32 rules in endomorphism type classes where only instances of valid rules are provided. So if the user is trying to perform an invalid operation like deriving a public key from a public key given a hard path, there will be no instance available for such an operation and the program will not compile (correct me if I got that wrong). 

My approach was to capture the bip32 rules in functions that are exposed to the users. The signature of the functions `derivePath` and `derivePubPath` define what is allowed and what is invalid. For instance, there is no legal application of one of those functions where you can get a public key from a public key using a hard path. I've put a lot of effort into having most of the types be derived automatically for the user so that he doesn't need to concern himself with them so much. For example, you can call:

```
let newPubKey = parsePubPath (Deriv :/ 0 :/ 1 :/ 2) somePubKey
```

and the system will automatically derive the type `SoftPath` for your derivation there. If you tried to do this:

```
let newPubKey = parsePubPath (Deriv :/ 0 :| 1 :/ 2) somePubKey
```

it would fail at compile time because the system will be unable to satisfy the types.

tphyahoo:
Thanks for listening.

I think we both understand the tradeoffs, so maybe it's time to step back, write more tests, and for me to pay more attention to haskoin-wallet and haskoin-faucet as a prototypical client library, which was my objective to begin with.

If the tests and application functionalities seem to want guarantees that the current system doesn't provide, I suppose we can revisit this.

Meanwhile I think more tests and build on top of #201 seems like the way to go.

tphyahoo:
Sorry if this is beating a dead horse, just a quick thought.

I believe a function that the current approach lacks is 

> deriveHard :: SomePathGuaranteedToBeHardWhichIsPossiblyMixed -> PrivKey -> PrivKey

 we have

> derivePath :: DerivPathI t -> XPrvKey -> XPrvKey

but that doesn't quite fit the task.  

I suppose there's a (maybe kludgy) way to achieve this by getting to a HardPath, but of course this doesn't support mixed paths.

The problem is that derivePath is too permissive

> derivePath (Deriv :/ 0 :/ 1 :: DerivPath) :: XPrvKey -> XPrvKey

Restating: the objective of hypothetical function deriveHard is to derive a private key whose public key is unguessable, guaranteed by the compile.  Using an arbitrary bip32 path, as long as it satisfies Bip32HardPath (at least one hard segment in mixed path).

I suppose in practice, with current approach, one would use 

> hardSubKey :: XPrvKey  -- ^ Extended Parent private key
>             -> KeyIndex -- ^ Child derivation index
>             -> XPrvKey  -- ^ Extended child private key
> hardSubKeys :: XPrvKey -> KeyIndex -> [(XPrvKey, KeyIndex)]

Note however that these use a KeyIndex as input, not a path, and certianly not a path with any kind of guarantee.

With the approach in current PR one could define 

> deriveHard :: Bip32HardPath -> PrvKey -> PrvKey
> deriveHard path prvk = ( appEndo . deriveXPrvKeyEndo $ path ) prvk

and it has the exact property I described

m0dev76:
> the objective of hypothetical function deriveHard is to derive a private key whose public key is unguessable, guaranteed by the compile

If I understand correctly, you mean a function that derives a private key `prv2` from a private key `prv1` but whose public key `pub2` can not be derived from the public key `pub1`. Because, in theory, once you have a private key `prv2` you can always derive the public key `pub2` from that private key (see function `derivePubKey`). What you may not be able to do is derive the public key `pub2` from the public key `pub1` if you have some hard path somewhere in the derivation.

So basically you would like to have a compiler-safe way of producing a hardened key `prv2`. I think that constraints of this kind should be encoded in types. The only type that comes close is the `HardPath` type but `HardPath` guarantees that _all_ the derivations are hard. In this new scenario, we would like a type that guarantees that _at least one_ derivation is hard (not all soft). Put another way, we want the opposite of the `SoftPath` type which guarantees that all derivations are soft.

There's also the `DerivPath` type but we can't use it because it is a homogeneous type that encompasses all types of derivations, so even _all soft_ derivations can fit in there.

So we actually would need a new kind of type that guarantees that _at least one_ derivation in a path is hard. If we have such a type (for example, called `OneHardPath`) we could simply reuse the existing `derivePath` function. I think this would be the correct approach to this problem. 

One simple way of implementing this is to provide a `toOneHardPath` function in the same spirit as `toSoft` and `toHard`. Such a function would act as a _guard_ and will only return a derivation of this type if some criteria are met.

```
toOneHardPath :: DerivPathI t -> Maybe OneHardPath
```

Any derivation path (expressed as `DerivPathI t`) could be expressed as a `OneHardPath` if the criteria are met and the function returns Just. 

Maybe there is another way of achieving this result by doing something more clever in the data type `DerivPathI`. Maybe we can have this information automatically inferred somehow in a similar way as `HardToGeneric` and `GenericOrSoft`. I need to think about it.

These are just some of my thoughts on the problem. I do agree with you that having a type that satisfies the properties of `OneHardPath` is useful and I think it would be a good addition to the library.

tphyahoo:
Yes, you understood me and I agree with what you are saying.

I wish we had better names though.

I think Bip32HardPath is better than OneHardPath, since it satisfies the Bip32 spec.

OneHard sounds like ExactlyOneHard.  Really it's AtLeastOneHard, if we name like that.

The current HardPath would be better named AllHardPath.  There is nothing in bip32 about needing all hard paths in a hard derivation, this is a haskoin-wallet specific thing.

But maybe it's too disruptive to rename.

m0dev76:
I'm of course open to renaming things such that they convey their intentions in a better way. I do prefer when the names are self-explanatory like your suggestion of `AllHardPath` rather than referring to external documentations like `Bip32HardPath`. I don't think that renaming the types is too disruptive. I would accept a pull request.

tphyahoo:
Also with the approach you outlined it sounds like you regain the SemiGroupoid (but not monoid) property for this kind of path, and that's a sign that things are moving solidly in the direction of correctness.

Since SemiGroupoids are basically addition but not identity, so they are usually list likes or string-likes, excluding empty strings.

tphyahoo:
renaming:

OK, how about 

HardPath -> AllHardPath

and the new at-least-one hard segment path is just HardPath.

A downside is it might be confusing to api users since it's an already existing type that now has a different meaning.

m0dev76:
OK that sounds good. I agree that there might be a problem with changing the meaning of an existing type. I guess that if we want to do it, the sooner the better. And we'd release it in 0.4 as it would be a backwards-incompatible change.

tphyahoo:
> Maybe there is another way of achieving this result by doing something more clever in the data type DerivPathI. Maybe we can have this information automatically inferred somehow in a similar way as HardToGeneric and GenericOrSoft. I need to think about it.

I have been trying to do this in branch https://github.com/tphyahoo/haskoin/tree/DerivPathII_experiment

What do you think of

```
data DerivPathII t where
    (::||)  :: EmptyOrAllHardII  t => !(DerivPathII t) -> !KeyIndex -> AllHardPathII
    (::|)   :: AnyPath t => !(DerivPathII t) -> !KeyIndex -> HardPathII
    (::/)   :: NonEmptyPath t        => !(DerivPathII t) -> !KeyIndex -> (DerivPathII t)
    (:://)  :: SoftII     t => !(DerivPathII t) -> !KeyIndex -> NonEmptySoftPathII
    DerivII :: EmptySoftPathII
```

::| and ::/ work almost as :| and :/ in DerivePathI, but capture more accurate type information, I think.

The difference is that ::/ needs to know whether it is appending to a hard path or a soft path, so it can inherit the right type if the next index is a soft index. (Soft if left path is soft, mixed hard otherwise.)

We retain the ability to create only-hard-segment hard paths, with the new :|| constructor.

:// is necessary to differentiate empty and non-empty soft paths, as well as guaranteeing a soft (non-mixed) path.

Fleshed out work-in-progress but compiling and not-breaking-anything, and with some examples, can be reviewed at:

https://github.com/tphyahoo/haskoin/blob/DerivPathII_experiment/haskoin-core/Network/Haskoin/Crypto/ExtendedKeys.hs

tphyahoo:
On second thought, in the light of day and with fresh eyes, I don't think this is ready for prime time yet.  

Rethinking...

tphyahoo:
I think this is better.

```
data SoftPathII = DerivII | (:://) SoftPathII KeyIndex
data HardPathII = (::|/) HardPathII KeyIndex | (::||) HardPathII KeyIndex | (::/|) SoftPathII KeyIndex

class ExtendableSoftII p where
    (~/) :: p -> KeyIndex -> p
class ExtendableHardII p where
    (~|) :: p -> KeyIndex -> HardPathII
instance ExtendableSoftII SoftPathII where
    (~/) = (:://) -- I wanted to use (:/) but : seems not allowed in class definition?
instance ExtendableSoftII HardPathII where
    (~/) = (::|/)
instance ExtendableHardII SoftPathII where
    (~|) = (::/|)
instance ExtendableHardII HardPathII where
    (~|) = (::||)


exampleSoftPath :: SoftPathII
exampleSoftPath = DerivII ~/ 1 ~/ 2
exampleHardPath1, exampleHardPath2, exampleHardPath3 :: HardPathII
exampleHardPath1 = exampleSoftPath ~| 3
exampleHardPath2 = DerivII ~| 1 
exampleHardPath3 = exampleHardPath2 ~/ 2 ~/ 3
```

This doesn't cover the scenario for AllHardPath which is exclusively hard indexes. But one, I don't think that derivation type is that important (it's a haskoin wallet specific thing, nothing in bip32 spec about it) and two, I think the above can be tweaked to include AllHardPath if I think a little  more.

tphyahoo:
I'm going to play around with replacing existing DerivPathI machinery with the above in branch

```
replace_DerivPathI_experiment
```

unless someone sees a problem.

I think this is going to improve the specificity and legibility of haskoin-wallet data model, and make it easy to make various improvements I have in mind.  But I'll have to come back with more about that later.

m0dev76:
>  I don't think that derivation type is that important (it's a haskoin wallet specific thing, nothing in bip32 spec about it) 

I'm was looking over the bip32 spec again and was wondering where this is covered? I know bip32 mostly defines the 3 types of child key derivations, the serialization formats and then goes into some high level details about suggested derivation trees and use cases. But they don't seem to suggest anything relating to typing of derivation paths, unless I've overlooked something? It feels to me that we've got a bit of flexibility on that end from the spec.

tphyahoo:
There's no problem with AllHardPath type per Bip32 spec, it just doesn't have any special security properties beyond vanilla (mixed) hard/soft paths, which is to say a path with at least one hard segment.

jprupp:
This is stale, closing. Reopen if necessary.

PR #204, opened 2015-12-08
Hard follow soft tphyahoo

tphyahoo:
builds on top of #201.

tests pass.  more tests.

some code cleanup.

introduces Bip32PathIndex data type that tags soft or hard indices, but in a less aggressive refactor than previously submitted #200.


tphyahoo:
Hope this can be pulled now that CI passes.

m0dev76:
Hey, thanks for this pull request. It looks all good to me, I'm happy to pull it in. Just waiting for the OK from @xenog.

tphyahoo:
@plaprade I noticed you got rid of 

> (++|) 

in https://github.com/tphyahoo/haskoin/commit/6950a11374cab872351525a2d3cd77c7736062df

Was this an oversight or intentional?

m0dev76:
Yes it was intentional. I thought about it for a while and I think that having only one concatenation function is plenty:

```
(++/) :: DerivPathI t1 -> DerivPathI t2 -> DerivPath
```

This function can concatenate two paths of any type (may they be hard, soft or generic) and return the result as a new path of type generic. If you know, for example, that you have concatenated two hard paths and that the result should be a hard path as well, you can simply pass your result through the `toHard` function to get a hard path.

I think its easier to have only one generic function that can concatenate anything. Otherwise, we have to start defining all sorts of different function that can concatenate soft + hard, hard + soft, hard + hard, soft + soft etc. This of course doesn't scale well if you start to introduce new types of derivations like `OneHardPath` that has been suggested in the discussion #200.

Another solution would have been to add a concatenation function of this form:

```
(++/) :: DerivPathI t -> DerivPathI t -> DerivPathI t
```

It basically says that concatenating two paths of some type `t` will yield a new path with the same type `t`. It is a bit more limiting however because you can only concatenate things of the same type (like hard -> hard -> hard or soft -> soft -> soft). But it is consistent. We could offer this function as a utility function which guarantees a bit of a stronger type checking when dealing with specialised derivation paths like hard or soft. I'm open to suggestions.

jprupp:
I think this pull request makes some improvements with respect to existing BIP32 code. We should merge it.

ACK

jprupp:
I think @plaprade's argument for a `(++/)` function that preserves phantom type information is good, if it can be done without adding too much complexity (I think it can).

PR #208, opened 2015-12-28
Return to 1MB blocks in Testnet3

jprupp:
Goes into a blockchain header download loop when connecting to 1MB-block hosts having a longer 8MB chain.


m0dev76:
ACK

PR #225, opened 2016-01-24
BIP-0111 compatibility

jprupp:


m0dev76:
For reference, the code in btcd that checks BIP0111 compatibility is:
https://github.com/btcsuite/btcd/blob/BTCD_0_12_0_BETA/peer/peer.go#L1064

``` go
func (p *Peer) isValidBIP0111(cmd string) bool {
    if p.Services()&wire.SFNodeBloom != wire.SFNodeBloom {
        if p.ProtocolVersion() >= wire.BIP0111Version {
            log.Debugf("%s sent an unsupported %s "+
                "request -- disconnecting", p, cmd)
            p.Disconnect()
        } else {
            log.Debugf("Ignoring %s request from %s -- bloom "+
                "support is disabled", cmd, p)
        }
        return false
    }

    return true
}
```

m0dev76:
Additionally, I think we should also check that the remote node can handle bloom filters before sending `filterload` messages. That is, the remote node has protocol version >= 70011 and the service bit for bloom filters is set.

PR #226, opened 2016-01-24
Corrected some JSON problems

jprupp:
I'm no longer exposing the thread id field.


m0dev76:
If you want to remove the thread id, you should probably also remove the code here:

https://github.com/haskoin/haskoin/blob/master/haskoin-node/Network/Haskoin/Node/Peer.hs#L147

Is there a reason why we should remove the thread id from debug information? It could be useful in some cases to correlate log messages.

PR #231, opened 2016-02-05
Submodules

jprupp:
Using submodules since `secp256k1-haskell` now uses submodules, and `stack` will not fetch submodules recursively when provided a Git entry. Also, this avoids specifying the Git entry for `secp256k1-haskell` in multiple `stack.yaml` files simultaneously for every update.


jprupp:
Closing to create single mammoth pull request for all changes I have been working on.

PR #242, opened 2016-02-14
documentation fix for applyPath

tphyahoo:
Small but legitimate haddock fix. 

This fails travis for "CABALVER=stack GHCVER=stack" configuration but I'm thinking the problem must be on the travis side as I don't see anything in the commit that would cause this. The other configurations pass.


jprupp:
I merged this but GitHub did not notice.

PR #243, opened 2016-02-14
applyPath tests

tphyahoo:
applyPath tests
s/derivePathVectors/applyPathVectors/
add applyPathVector originating with public key
testBadApplyPath (applyPath that should result in errors)


tphyahoo:
this has same travis issue for "CABALVER=stack GHCVER=stack" configuration as

https://github.com/haskoin/haskoin/pull/242

but it's spurious.

PR #245, opened 2016-02-15
Best block height shown in status

jprupp:


m0dev76:
I ACK this pull request but I wanted to clarify the behaviour here for posterity. The reported block hash and block height here is an approximation. The node (the STM code) doesn't really know at what height the wallet really is. It can guess those values when the wallet requests a block download from a given block hash for example. If we wanted to report the correct value (block hash and height with respect to the wallet) we would have to look in the wallet database and report the values there directly. But haskoin-node doesn't have access to the wallet database.

PR #246, opened 2016-02-15
Removed pagination from API

jprupp:


m0dev76:
ACK. This will make the API nicer to use. I wonder who decided to use pages in the first place!

PR #247, opened 2016-02-15
Rebroadcast all

jprupp:
Do not limit the number of transactions to be rebroadcast to avoid broadcasting partial pending transaction chains.


m0dev76:
ACK. Although I am wondering if we should not implement some sort of exponential backoff rebroadcasting (rebroadcast an unconfirmed transaction when the next block arrives, then in 2 blocks, then in 4, then in 8 etc.)

jprupp:
Blocks arrive every ten minutes in production. That is not too often. If a transaction misses the next block it is reasonable to retransmit. In any case we are only sending INV messages to our peers, so bandwidth consumption is negligible.

PR #248, opened 2016-02-15
Fix delete TX

jprupp:


haskoin-wallet/Network/Haskoin/Wallet/Transaction.hs
```diff
@@ -223,15 +223,18 @@ addrTxPage (Entity ai _) addrI page@PageRequest{..}
...
-        where_ $ t ^. KeyRingTxHash ==. val txid
-        limit 1
-        return $ t ^. KeyRingTxTx
-    case res of
-        (Value tx:_) -> return $ Just tx
-        _ -> return Nothing
+getTx :: MonadIO m => TxHash -> SqlPersistT m (Maybe (Entity KeyRingTx))
+getTx txid = fmap listToMaybe $ select $ from $ \t -> do
+    where_ $ t ^. KeyRingTxHash ==. val txid
+    limit 1
```
m0dev76: This query will only return the first transaction. The same transaction may exist in multiple accounts (especially when building offline transactions for example) so this may not be sufficient to delete all occurrences of a given TxID.

haskoin-wallet/Network/Haskoin/Wallet/Server/Handler.hs
```diff
@@ -522,14 +522,16 @@ getTxR keyRingName name txid = do
...
             => TxHash -> Handler m (Maybe Value)
 deleteTxIdR txid = do
     $(logInfo) $ format $ unlines
         [ "DeleteTxR"
         , "  TxId: " ++ cs (txHashToHex txid)
         ]
-    runDB $ deleteTxId txid
+    runDB $ do
+        tM <- getTx txid
+        maybe (return ()) deleteTx tM
```
m0dev76: Here we only delete the first occurrence of `txid`. See my other comment.

jprupp:
I noticed this and I have a fix in the devel branch. I will backport it to master.

PR #254, opened 2016-02-21
Simplify API

jprupp:
Removes the `JsonWith*` data types for simplicity.

### DIFF
--- a/haskoin-wallet/Network/Haskoin/Wallet/Transaction.hs
+++ b/haskoin-wallet/Network/Haskoin/Wallet/Transaction.hs
@@ -47,7 +47,7 @@ import Data.Either (rights)
 import Data.List ((\\), nub, nubBy, find)
 import Data.Text (unpack)
 import Data.Conduit (Source, mapOutput)
-import Data.Maybe (isNothing, isJust, fromMaybe, listToMaybe)
+import Data.Maybe (isNothing, isJust, fromMaybe, listToMaybe, catMaybes)
 import qualified Data.Map.Strict as M
     ( Map, toList, map, unionWith, fromListWith )
 import Data.String.Conversions (cs)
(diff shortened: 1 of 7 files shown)


haskoin-wallet/Network/Haskoin/Wallet/Transaction.hs
```diff
@@ -173,7 +174,7 @@ addrTxs (Entity ai _) addrI ListRequest{..} = do
         tidList = fOrd $ take (fromIntegral listLimit) $
             drop (fromIntegral listOffset) $ fOrd tids
 
-    -- Use a sliptSelect query here with the exact tids to speed up the
+    -- Use a splitSelect query here with the exact tids to speed up the
     -- query.
     res <- splitSelect tidList $ \tid ->
```
m0dev76, line 179: I think that this query can be optimized as we don't need the aggregates. We can probably get rid of the left joins and the group by clause.

jprupp, line 179: There is a future commit that addresses precisely this in the devel branch, but it doesn't apply cleanly against master.

> @@ -173,7 +174,7 @@ addrTxs (Entity ai _) addrI ListRequest{..} = do
>          tidList = fOrd $ take (fromIntegral listLimit) $
>              drop (fromIntegral listOffset) $ fOrd tids
> -    -- Use a sliptSelect query here with the exact tids to speed up the
> -    -- Use a splitSelect query here with the exact tids to speed up the
>    -- query.
>    res <- splitSelect tidList $ \tid ->

I think that this query can be optimized as we don't need the aggregates. We can probably get rid of the left joins and the group by clause. 

---

Reply to this email directly or view it on GitHub:
https://github.com/haskoin/haskoin/pull/254/files#r53884249

m0dev76:
Okay then we can merge this in and take care of the devel branch when its ready.

Issue #214, opened 2016-01-10
Send transactions with time locked funds (BIP-65)


jprupp:
A new transaction command, under the "transactions" section in help.

sendtimelocked acc addr amount unlockBlock [-S]

funds are spendable after unlockBlock appears in the blockchain.

tphyahoo:
Is anyone working on this?

If not I may begin (I have probable use for it).  My first move would be to wrangle bip65 test vectors from

https://github.com/petertodd/bitcoin/commit/ab0f54f38e08ee1e50ff72f801680ee84d0f1bf4 (linked from bip65 spec)

into 

https://github.com/haskoin/haskoin/blob/master/haskoin-core/tests/Network/Haskoin/Transaction/Units.hs

which seems to be basically cribbed from satoshi bitcoin tx_valid.json and tx_invalid.json

UPDATE: I suppose it would be better to use vectors repo bitcoin/bitcoin instead of petertodd/bitcoin, as more up to date.  

Also, it's a pity we can't just copy tx_valid.json and tx_invalid.json (and many other test data files there as well potentialy) from bitcoin repo and use these for tests directly, rather than staying in sync by manual surgery.  Perhaps there would be some way to achieve that.

tphyahoo:
transaction tests now include failing vectors for cltv, copied from bitcoin core test suite (where the vectors pass). (from #255)

jprupp:
Script interpreter code is outdated. Work needs to be done update it or remove it from `haskoin-core`. I'll close this bug because it is just one part of changes that should be done to the script interpreter.

PR #255, opened 2016-02-21
Parsed satoshi core txvalidjson tests

tphyahoo:
use tx_valid.json from bitcoin core, parse it for same style of transaction tests that were already running.

65 / 83 tests fail. 

I haven't done an assessment of all the various fail cases, and probably some are spurious. Hope to do more on that soon.

I'm confident that there are (at least) csv/cltv related fails here, that can be used as a starting point for #214.  So I think it's worth merging for that alone.


tphyahoo:
Is it permitted to merge failing tests to master? 

The easiest way to get all passing is to revert the changes to Main.hs in test suite (which just skips the tests for now).

jprupp:
Skip the tests that are failing and add TODOs to re-enable when fixed. The test suite must always pass.

haskoin-core/Network/Haskoin/Script/Types.hs
```diff
@@ -53,7 +53,7 @@ instance Binary Script where
             empty <- isEmpty
             if empty
                 then return []
-                else liftM2 (:) get getScriptOps
+                else (pure (:)) <*> get <*> getScriptOps
```
jprupp: I think this is the same as:

``` haskell
(:) <$> get <*> getScriptOps
```

tphyahoo: yes, it's precisely the same.

tphyahoo:
Test suite now passes, with

```
+    passingTests = [0..5] ++ [8] ++ [11..13] ++ [16..18] ++ [20] ++ [52]
+    isCurrentlyPassing (_, testNum) = elem testNum passingTests
```

15 new passing tests.

haskoin-core/Network/Haskoin/Script/Types.hs
jprupp: Can you change it for the simpler version using `(<$>)`? I will merge when you do and @plaprade gives his ack. Consider this pull request "acked" by me.

tphyahoo: ok I made the change: fcbf97b

tphyahoo:
I like pure f <*> instead of f <$> (which are identical) because it seems more symmetrical to me, and the translation from pure

```
f a b c
```

to 

```
pure f <*> pure a <*> pure b <*> pure c
```

rather than

```
f <$> pure a <*> pure b <*> pure c
```

seems more symmetrical to me.

But, the <$> variant sees to be more prevalent in the community.

Of course, it doesn't really matter since both are totally equivalent.

Just 2c.

jprupp:
I have merged this as a single squashed commit. GitHub does not notice that the code has been merged.

PR #262, opened 2016-02-27
Rename account

jprupp:


m0dev76:
I think you forgot to add the rename command to the Client.hs file

jprupp:
Yeah, possibly lost in one of the rebasings. I'll add it when I get to a computer.

m0dev76:
ACK

PR #265, opened 2016-03-01
Parsed satoshi core txvalidjson tests

tphyahoo:
parses test description in json, gives better feedback in tests.

PR #266, opened 2016-03-02
instrument bitcoin core json tests vectors better (include comments

tphyahoo:
This replaces #265 which I abandoned due to either git bug or personal fail. 

It improves test instrumentation, by including comments from the bitcoin core json test vectors in the test output.


tphyahoo:
I am trying again on #265, now that I understand merge process better, so pausing here.

PR #268, opened 2016-03-03
Parsed satoshi core txvalidjson tests 2

tphyahoo:
This replaces #265 which I abandoned due to either git bug or personal fail.

It improves test instrumentation, by including comments from the bitcoin core json test vectors in the test output.

I also cleaned up the json parsing a bit.


jprupp:
Manually merged already.

PR #271, opened 2016-03-22
Moving HeaderTree to SQL

jprupp:
Also added unit tests to HeaderTree, and fixed a bug while two SIGINT calls were needed to kill hw (Ctrl-C twice).


m0dev76:
ACK

PR #278, opened 2016-05-11
Prune chain to avoid too many forks

jprupp:


jprupp:
Meh, I merged GitHub-style, my bad.

PR #281, opened 2016-05-29
ØMQ fix

jprupp:
Fix ØMQ sockets to avoid them keeping processes or threads alive indefinitely.


m0dev76:
ACK, I'll add those changes into the address-index branch as well

PR #288, opened 2016-10-04
Add OP_RETURN ScriptOutput

hudon:


jprupp (approved):
Great, thank you for this. Very useful.

hudon:
@xenog Thanks for merging! I had one question: Core allows 83 bytes (including the OP_RETURN and any PUSHDATA opcodes) but I'm not sure where that would be enforced in Haskoin? I'd be glad to add that in too if you have an idea of where it would go... `encodeOutput`?

jprupp:
@hudon I haven’t thought about it. It would be cool if the output can only be created when it fits. Sometimes my approach to this is to not export the constructors and only expose functions that do validation to construct the data. I’m open to suggestions.

PR #291, opened 2016-10-22
Changes required to use "inproc" ZeroMQ transport + example app

runeksvendsen:
First commits adds `runSPVServerWithContext` (by separating it out from `runSPVServer`), which allows passing a specific ZeroMQ `Context` to the haskoin-wallet server. A shared Context between receiver and sender is required in order to speak to the server in-process/inter-thread via the "inproc" ZeroMQ transport.

Second commit adds an example executable that embeds a haskoin-wallet server in-app, using `runSPVServerWithContext` and the "inproc" transport for inter-thread communication with the server. (If I had more ncurses skills I would create a proper GUI for it, but for now it makes most sense to run it as `example-inproc-wallet-server 2> debug.log` and, optionally, view the log using `tail -f debug.log`)

### DIFF
--- a/haskoin-core/Network/Haskoin/Crypto/ECDSA.hs
+++ b/haskoin-core/Network/Haskoin/Crypto/ECDSA.hs
@@ -91,11 +91,11 @@ hashToMsg =
 
 -- <http://www.secg.org/sec1-v2.pdf Section 4.1.3>
 -- | Sign a message
-signMsg :: Hash256 -> PrvKey -> Signature
+signMsg :: Hash256 -> PrvKeyI c -> Signature
 signMsg h d = Signature $ EC.signMsg (prvKeySecKey d) (hashToMsg h)
 
 -- | Verify an ECDSA signature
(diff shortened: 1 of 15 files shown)


runeksvendsen:
In a completely unrelated move, I've also added a `pubkeys` command to the `hw` command line utility, for displaying the public key related to an address.

runeksvendsen:
Question: is the only case where a `WalletAddr` _doesn't_ contain a compressed pubkey (in the "key" field) the case where the account is multisig? Or would a regular account (read-only or not) have some reason not to contain a `PubKeyC` in the `key` field for a `WalletAddr`?

jprupp:
Looks really cool. I will wait and see what @plaprade has to say about it. I haven’t tested yet but I like the direction this is taking.

runeksvendsen:
I have factored out the "inproc"/runSPVServerWithContext patch into a separate branch. So I'm closing this pull request, and will open two separate ones.

PR #301, opened 2017-08-06
Generalize the display of public keys

m0dev76:
At the command line, public keys can now be requested instead of
addresses (for relevant commands) by using the -k or --pubkeys option.
This generalizes the previous pubkeys command which has now been
removed.

This commit also includes a bit of cleanup and style changes in the
wallet code.


xenog:
ACK

jprupp (approved):

PR #303, opened 2017-08-10
Generate mnemonics with up to 256 bits of entropy

m0dev76:
This patch introduces the possibility to generate mnemonics using up to
256 bits (32 bytes) of entropy. The default entropy of 128 bits (16
bytes) can be changed using the command line option -e (or --entropy).
You must pass the entropy value in bytes.  Valid values for the
--entropy option are 16, 20, 24, 28 and 32. Using more entropy will
yield longer mnemonics.

Key derivations for account master keys are now computed server side instead of in the command line (if a custom derivation was requested). This simplifies the zmq API logic a bit.

There's also a bit of code refactoring at the account creation level.

### DIFF
--- a/haskoin-wallet/Network/Haskoin/Wallet/Accounts.hs
+++ b/haskoin-wallet/Network/Haskoin/Wallet/Accounts.hs
@@ -46,11 +46,13 @@ import Control.Monad.Trans.Resource (MonadResource)
 import Control.Exception (throw)
 
 import Data.Text (Text, unpack)
-import Data.Maybe (mapMaybe, listToMaybe, isJust, isNothing)
+import Data.Default (def)
+import Data.Maybe
+    (mapMaybe, listToMaybe, maybeToList, isJust, isNothing, fromMaybe)
 import Data.Time.Clock (getCurrentTime)
 import Data.Time.Clock.POSIX (utcTimeToPOSIXSeconds)
 import Data.List (nub)
-import Data.Word (Word32)
+import Data.Word (Word8, Word32)
 import Data.String.Conversions (cs)
 import Data.Serialize (encode)
 
@@ -70,12 +72,12 @@ import Network.Haskoin.Crypto
 import Network.Haskoin.Block
 import Network.Haskoin.Script
 import Network.Haskoin.Node
-import Network.Haskoin.Util
 import Network.Haskoin.Constants
 import Network.Haskoin.Node.HeaderTree
 
 import Network.Haskoin.Wallet.Types
 import Network.Haskoin.Wallet.Model
+import Network.Haskoin.Wallet.Settings
 
 {- Initialization -}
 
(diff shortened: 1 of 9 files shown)


haskoin-wallet/Network/Haskoin/Wallet/Client/Commands.hs
```diff
@@ -154,26 +154,32 @@ getKey = do
...
         , Nothing
         , derivM <|> Just defaultDeriv
         , Nothing
         )
     go str' derivM = case xPrvImport str' of
+        -- Provided account private key. No further derivation should be
+        -- required unless one was given.
         Just k -> return
             ( Nothing
-            , Just $ maybe k (`derivePath` k) derivM
```
m0dev76, line 166: Need to check if all the logic is still correct here. Instead of doing one odd derivation client side, all derivations are now computed server side with the given derivation path.

jprupp: That line proves beyond any doubt that I am perfectly capable to write ugly incomprehensible bad quality code, even in Haskell, which I would consider an achievement.

haskoin-wallet/Network/Haskoin/Wallet/Accounts.hs
```diff
@@ -124,56 +126,63 @@ initGap accE = do
...
+                        -- Generate a key if nothing was provided
+                        let defEnt = configEntropy def
+                        ms   <- genMnemonic $ fromMaybe defEnt newAccountEntropy
+                        root <- mnemonicToPrvKey ms
+                        return (Just ms, Just root)
+                    else return (Nothing, Nothing)
+
+    let masterM = case newAccountDeriv of
+                      Nothing -> rootM
+                      Just d  -> (derivePath d) <$> rootM
```
m0dev76, line 174: Derivations are computer server side here if a path was provided in the API call. I think this makes the API a bit more general.

haskoin-wallet/Network/Haskoin/Wallet/Accounts.hs
```diff
@@ -124,56 +126,63 @@ initGap accE = do
     void $ createAddrs accE AddressExternal 20
     void $ createAddrs accE AddressInternal 20
 
--- | Create a new account
+genMnemonic :: (MonadThrow m, MonadIO m) => Word8 -> m Mnemonic
+genMnemonic bytes
+    | bytes `elem` [16..32] = do
```
jprupp: `[16,20..32]`

haskoin-wallet/Network/Haskoin/Wallet/Accounts.hs
```diff
@@ -124,56 +126,63 @@ initGap accE = do
...
+        entropy <- liftIO $ getEntropy $ fromIntegral bytes
+        case toMnemonic entropy of
+            Right ms -> return ms
+            Left err -> throwM $ WalletException err
+    | otherwise = throwM $
+        WalletException "Entropy can only be between 16 and 32 bytes"
```
jprupp: ...and divisible by 4.

haskoin-wallet/Network/Haskoin/Wallet/Client/Commands.hs
jprupp: This was the last of the client-side derivations that I didn't remove in a previous refactor. Good catch.

haskoin-wallet/Network/Haskoin/Wallet/Client/Commands.hs
```diff
@@ -183,19 +189,20 @@ getKey = do
 
 -- First argument: is account read-only?
 cmdNewAcc :: Bool -> String -> [String] -> Handler ()
-cmdNewAcc r name ls = do
+cmdNewAcc readOnly name ls = do
+    e <- R.asks configEntropy
     _ <- return $! typ
-    e <- checkExists name
-    when e $ error "Account exists"
+    checkExists name >>= flip when (error "Account exists")
```
jprupp:
```
(`when` error "Account exists")
```

jprupp (changes requested):

jprupp (approved):

PR #309, opened 2017-08-18
Payment requests and offline signing improvements

m0dev76:
Add the uri command to the command line tool allowing users to generate
Bitcoin payment requests. The output of this command can be piped to qr
code generators like qrencode.

The getoffline command has been changed to print a base64 blob
containing the transaction and coin signing data. The signoffline
command will now read this base64 blob from stdin. The output of
getoffline can be piped into signoffline. These changes allow you to use
qr codes to pass data between these commands.

### DIFF
--- a/haskoin-wallet/Network/Haskoin/Wallet/Client/Commands.hs
+++ b/haskoin-wallet/Network/Haskoin/Wallet/Client/Commands.hs
@@ -10,6 +10,7 @@ module Network.Haskoin.Wallet.Client.Commands
 , cmdList
 , cmdUnused
 , cmdLabel
+, cmdURI
 , cmdTxs
 , cmdAddrTxs
 , cmdGetIndex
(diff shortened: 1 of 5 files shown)


haskoin-wallet/Network/Haskoin/Wallet/Client/Commands.hs
```diff
@@ -358,6 +361,31 @@ cmdGenAddrs name i = do
...
+buildPaymentRequest :: Address -> [String] -> BS.ByteString
+buildPaymentRequest a ls =
+    "bitcoin:" <> addrToBase58 a <> cs params
+  where
+    params = if null paramStr then "" else "?" <> paramStr
+    paramStr = concat $ intersperse "&" $ zipWith ($)
+        [ ("amount" `buildParam`) . parseAmount
+        , ("message" `buildParam`) . URI.encode
+        ] ls
+    parseAmount str = show (read str :: Float)
```
jprupp: Use a `Double` here since `Float` loses precision no large Bitcoin amounts if considering satoshi.

haskoin-wallet/Network/Haskoin/Wallet/Types.hs
```diff
@@ -233,13 +238,44 @@ data CoinSignData = CoinSignData
 
 $(deriveJSON (dropFieldLabel 8) ''CoinSignData)
 
+instance Serialize CoinSignData where
+    get = do
+        p <- get
+        s <- readBS >>= \bs -> case decodeOutputBS bs of
+            Right s -> return s
+            _ -> error "Invalid ScriptOutput in CoinSignData"
+        readBS >>= \bs -> case Aeson.decode $ BL.fromStrict bs of
```
jprupp: No.

haskoin-wallet/Network/Haskoin/Wallet/Types.hs
```diff
@@ -233,13 +238,44 @@ data CoinSignData = CoinSignData
...
+            Just d -> return $ CoinSignData p s d
+            _ -> error "Could not decode SoftPath in CoinSignData"
+      where
+        readBS = get >>= \(VarInt l) -> getByteString $ fromIntegral l
+
+    put (CoinSignData p s d) = do
+        put p
+        writeBS $ encodeOutputBS s
+        writeBS $ BL.toStrict $ Aeson.encode d
```
jprupp: Ditto.

jprupp (approved):

jprupp (approved):

PR #317, opened 2017-08-28
Better arbitrary instances in haskoin-core

m0dev76:
Instead of having arbitrary types for testing, we now provide arbitrary
functions of type `Gen a`. This simplifies the process of writing new
arbitrary instances for tests in other libraries like haskoin-wallet.

### DIFF
--- a/haskoin-core/Network/Haskoin/Crypto/ExtendedKeys.hs
+++ b/haskoin-core/Network/Haskoin/Crypto/ExtendedKeys.hs
@@ -74,7 +74,7 @@ import           Data.Aeson                    (FromJSON, ToJSON,
                                                 toJSON, withText)
 import           Data.Bits                     (clearBit, setBit, testBit)
 import           Data.ByteString               (ByteString)
-import qualified Data.ByteString               as BS (append, take)
+import qualified Data.ByteString               as BS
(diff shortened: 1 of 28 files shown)


jprupp (approved):
Much nicer.

runeksvendsen:
Why not keep the old `Arbitrary` types, and add the functions as an addition? I have quite a lot of code that depends on the types, and I see no reason to throw these out, thus breaking a lot of code that depends on these. I think it would make more sense to just add the functions and keep the `ArbitraryX` types.

runeksvendsen:
On the other hand it does reduce duplication not having to write e.g. `ArbitraryPubKey pubkey <- arbitrary` but rather `pubkey <- arbitraryPubKey`. So I think it makes sense if we bump the major version to denote the breaking change.

jprupp:
Sorry I merged before I saw your message. It didn't take long to refactor all tests with these changes. It is indeed less repetitive. If you need any help shout.

runeksvendsen:
No problem. I think we all agree that it’s a cleaner interface, and if we keep the naming conventions, except starting with a lowercase letter, it should indeed be a quick refactor. But I also assume we agree that backwards incompatible changes to the `Test` module warrant a major version bump.

PR #318, opened 2017-08-31
Move files

jprupp:
- Move files in line with what the Stack tool uses as default directory structure for new packages.
- Reserve the `Network.Haskoin.Node` namespace for the `haskoin-node` package (future additions in pipeline).


m0dev76 (commented):
Fine, I'll humor your ocd :)

PR #320, opened 2017-08-31
Use ShortByteString and cryptonite for hashes

jprupp:
- Must set network before using constants (Prodnet no longer default).
- Reduce memory fragmentation by using ShortByteString for hashes.
- Use cryptonite instead of deprecated cryptohash for hashes.
- Add Regtest network support.
- Bump secp256k1 library version.
- Improve encodeCompact and decodeCompact functions.
- Refactor visited files.
- Add extra data to network constants.
- Change serialisation functions.
- Add sendheaders message.
- Remove transaction and block hash from type.
- Additional minor changes.

### DIFF
--- a/haskoin-core/haskoin-core.cabal
+++ b/haskoin-core/haskoin-core.cabal
@@ -91,9 +91,11 @@ library
                  , cereal                   >= 0.5          && < 0.6
                  , conduit                  >= 1.2          && < 1.3
                  , containers               >= 0.5          && < 0.6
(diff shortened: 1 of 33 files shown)


runeksvendsen:
Are you able to separate the changes introduced by `hlint` from the rest of the changes? As far as I can see, a lot of the changes are suggestions by `hlint`, which don’t seem necessary to check manually.

jprupp:
I'm withdrawing this pull request and coming up with a bunch of smaller ones.

PR #321, opened 2017-09-07
Upgrading to GHC 8.2.1

wraithm:


wraithm:
@xenog, would you mind merging this?

runeksvendsen:
I'm unable to build and test this PR using GHC 8.2.1 because `secp256k1` depends on an old Cabal version in its `custom-setup` stanza. From its cabal file:

```
custom-setup
    setup-depends:       base >= 4.8 && < 5
                       , Cabal >= 1.24 && < 1.25
```

I'm not sure whether it's sufficient to just bump this Cabal version range for `secp256k1` as well, so I thought I'd write here instead. Perhaps this is all it takes, but I don't have any experience with this stuff. Care to take a look at this, @xenog?

wraithm:
So, I bumped the `Cabal` package upper bound version to `< 2.1`, and `secp256k1` seems to compile and work fine. The test suite passes at least. I could submit a PR to `secp256k1` with the version upper bound bump. I'm not that familiar with the PVP, but it feels wrong to me to have the version bounds be `>= 1.24 && < 2.1`... What do you guys think is the best course of action?

wraithm:
Actually, I read the PVP. It doesn't say much about dependency version changes. It really only talks about the exposed API. So, I'm going to make that PR to `secp256k1`.

wraithm:
https://github.com/haskoin/secp256k1-haskell/pull/7

runeksvendsen:
In my view, as long as `secp256k1` builds fine, there shouldn’t be any issues, since — as far as I know — the `custom-setup` stanza is only relevant to the build process, and is not a dependency of the library itself.

Could you try running the tests for Haskoin, using GHC 8.2.1 with a custom `stack.yaml` pointing to the `secp256k1` that contains the Cabal upper-range-version bump? I’d like to see whether or not all Haskoin tests finish without errors using GHC 8.2.1 (plus the new dependency versions), but I don’t have the time right now myself.

wraithm:
So, I tried to use `secp256k1`'s master version (0.5.2). `haskoin-core` is incompatible with the new version of `secp256k1`. Specifically, the tests don't compile. I'm going to try to upgrade the old version of `secp256k1` (0.4.8) to use the new Cabal and see what happens. I might try to upgrade `haskoin-core` to use `secp256k1` 0.5.2 if that fails.

wraithm:
Okay, this version works with `secp256k1` 4.8 if you bump the `Cabal` version. All the tests pass on `haskoin-core`.

wraithm:
`secp256k1` master has already upgraded Cabal versions, but it doesn't work with this version of `haskoin-core`.

jprupp:
Everything has been updated. I pushed a barrage of changes during the last few days. We are now at the latest LTS Haskell.

wraithm:
Thanks @xenog!

PR #323, opened 2017-09-25
BIP 144 - serialization/deserialization of transaction

t4sk:
This PR makes several modifications across `haskoin` in order to support serialization / de-serialization of transaction defined in BIP 144.

The changes are briefly summarized below:
- Add new serialization function for transactions with witness data
- Add new datatype Witness and store it in TxIn
- JSON encoder for witness data

Here is an example of a transaction taken from [BIP 143](https://github.com/bitcoin/bips/blob/master/bip-0143.mediawiki):
```
01000000000102fff7f7881a8099afa6940d42d1e7f6362bec38171ea3edf433541db4e4ad969f00000000494830450221008b9d1dc26ba6a9cb62127b02742fa9d754cd3bebf337f7a55d114c8e5cdd30be022040529b194ba3f9281a99f2b1c0a19c0489bc22ede944ccf4ecbab4cc618ef3ed01eeffffffef51e1b804cc89d182d279655c3aa89e815b1b309fe287d9b2b55d57b90ec68a0100000000ffffffff02202cb206000000001976a9148280b37df378db99f66f85c95a783a76ac7a6d5988ac9093510d000000001976a9143bde42dbee7e4dbe6a21b2d50ce2f0167faa815988ac000247304402203609e17b84f6a7d30c80bfa610b5b4542f32a8a0d5447a12fb1366d7f01cc44a0220573a954c4518331561406f90300e8f3358f51928d43c212a8caed02de67eebee0121025476c2e83188368da1ff3e292e7acafcdb3566bb0ad253f62fc70f07aeee635711000000
```
The serialization above can be quickly checked with `hw decode`. In the output there is a new field called `witness` inside each input.

### DIFF
--- a/haskoin-core/src/Network/Haskoin/Test.hs
+++ b/haskoin-core/src/Network/Haskoin/Test.hs
@@ -92,6 +92,7 @@ module Network.Haskoin.Test
 , arbitraryTxHash
 , arbitraryTxIn
 , arbitraryTxOut
+, arbitraryWitness
 , arbitraryOutPoint
 , arbitraryAddrOnlyTx
 , arbitraryAddrOnlyTxIn
--- a/haskoin-core/src/Network/Haskoin/Test/Transaction.hs
+++ b/haskoin-core/src/Network/Haskoin/Test/Transaction.hs
@@ -12,6 +12,7 @@ import           Network.Haskoin.Crypto
 import           Network.Haskoin.Script
 import           Network.Haskoin.Test.Crypto
 import           Network.Haskoin.Test.Script
+import           Network.Haskoin.Test.Util
 import           Network.Haskoin.Transaction
 import           Network.Haskoin.Util
 import           Test.QuickCheck
@@ -39,11 +40,27 @@ arbitraryTxOut =
     TxOut <$> (getTestCoin <$> arbitrarySatoshi)
           <*> (encodeOutputBS <$> arbitraryScriptOutput)
 
--- | Arbitrary TxIn
+-- | Arbitrary Witness
+arbitraryWitness :: Gen Witness
+arbitraryWitness = do
+    n <- choose (0, 5)
+    ws <- vectorOf n arbitraryBS
+    return $ Witness ws
+
+-- | Arbitrary TxIn without Witness
 arbitraryTxIn :: Gen TxIn
 arbitraryTxIn =
     TxIn <$> arbitraryOutPoint
          <*> (encodeInputBS <$> arbitraryScriptInput)
+         <*> return (Witness [])
+         <*> arbitrary
+
+-- | Arbitrary TxIn with Witness
+arbitraryTxInWitness :: Gen TxIn
+arbitraryTxInWitness =
+    TxIn <$> arbitraryOutPoint
+         <*> (encodeInputBS <$> arbitraryScriptInput)
+         <*> arbitraryWitness
          <*> arbitrary
 
 -- | Arbitrary Tx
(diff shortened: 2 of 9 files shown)


runeksvendsen:
Hi there

First of all: thank you for writing the code for this!

Secondly, here are some thoughts on the subject, which I'd like to discuss:

* Would it make sense to create a new transaction type, e.g. `WitTx` which can be serialized from/to transactions containing SegWit inputs? The rationalization for this is that `Tx` is an existing, well-tested data structure, and I'd prefer to keep this as-is, and add `WitTx` as a wrapper around `Tx` which adds the relevant witnesses for SegWit transactions. I imagine `WitTx` would wrap a regular `Tx`, plus a `Map Word32 Witness`, which maps zero or more input indexes to a `Witness`. So, a non-SegWit transaction would be a `WitTx` with an empty map (no witnesses), whereas a SegWit-transaction would comprise a `WitTx` one or more entries in the "witness map". I'm honestly not sure whether this is the right way to go, but I'd like to throw it out there as a suggestion. I feel like it makes more sense to implement SegWit as an additional layer, as opposed to modifying the core of an already well-tested -- and seemingly robust -- implementation.

* Tests. I think we need to create a parser for the various SegWit-related test vectors (valid/invalid SegWit tx's) contained in the Bitcoin Core source code, such that we are able to run tests for Haskoin's SegWit implementation against these test vectors.

t4sk:
Hi @runeksvendsen 

Thanks for the feedback!
I agree that creating `WitTx` is a safer approach, so I will create another PR.

jprupp:
Hi,

I am going to throw my opinion here. I wrote the SegWit implementation for bitcoinJ, so I am quite familiar with implementation quirks. I would go for reusing the existing `Tx` data structure, as transactions are now all witness transactions, and they have an implicit witness structure. It makes sense to add this code to the existing structure to reflect this, even when there are two different ways to serialise transactions now.

For serialisation, the `Serialize` instance for transactions should serialise by default with witness data, as in the new serialisation format. The `txHash` function should return only the transaction id obtained from the serialisation of the transaction excluding witness data. There should be an extra function available to allow serialisation without witness data when it is required by peers that do not set the witness flag.

What do you think?

jprupp:
I reopened this pull request, although it isn't necessary to merge yet. Let's have a discussion.

t4sk:
Hi @xenog,

Thanks for your feedback.

This PR took the approach suggested by you.
Perhaps we can improve on this PR and test it on a new branch like `haskoin:segwit`?

I am happy to setup a `haskoin` server / wallet on testnet

runeksvendsen:
I’m not opposed to doing it this way at all. In fact, it may be the right way to do it. I just threw my thoughts out there. 

The way I suggested would leave us with two transaction data structures, which wouldn’t be elegant at all. Perhaps my opinion is and old one — from a time where SegWit was just a proposal, i.e. it hadn’t activated yet — but now that it has activated, and is part of the Bitcoin protocol, I think it makes sense to replace the old `Tx` data structure, and make it SegWit-compatible.

Also, mind you, I have only read the SegWit-related BIPs (I haven’t actually implemented it), so I think it makes sense to go with the approach in this PR, and as suggested by you, @xenog.

jprupp:
Great idea, make it so.

runeksvendsen:
@t4sk I’ve created a new “segwit” branch, can you recreate both of your pull requests against this branch?

t4sk:
@runeksvendsen Thanks for creating the branch.

[Here](https://github.com/haskoin/haskoin/pull/329) is the new pull request for this one

runeksvendsen:
Merged into `segwit` branch.

PR #335, opened 2017-12-28
Improve compact encoding/decoding

jprupp:
Compact numbers can also be negative. Code rewritten for better compatibility with reference implementation.

### DIFF
--- a/haskoin-core/src/Network/Haskoin/Block/Types.hs
+++ b/haskoin-core/src/Network/Haskoin/Block/Types.hs
@@ -362,10 +362,8 @@ encodeCompact i = nCompact
     i' = abs i
     neg = i < 0
     nSize' :: Int
-    nSize' = let f n = let n' = n `shiftR` 8
-                       in if n' == 0
-                          then 1
-                          else 1 + f n'
+    nSize' = let f 0 = 0
+                 f n = 1 + f (n `shiftR` 8)
              in f i'
     nCompact''' :: Word32
     nCompact'''
--- a/haskoin-core/src/Network/Haskoin/Block/Types.hs
+++ b/haskoin-core/src/Network/Haskoin/Block/Types.hs
@@ -32,7 +32,7 @@ import           Data.Aeson                        (FromJSON, ToJSON,
 import           Data.Bits                         (shiftL, shiftR, (.&.),
                                                     (.|.))
 import           Data.ByteString                   (ByteString)
-import qualified Data.ByteString                   as BS (length, reverse)
+import qualified Data.ByteString                   as BS (reverse)
 import           Data.Maybe                        (fromMaybe)
 import           Data.Serialize                    (Serialize, encode, get, put)
 import           Data.Serialize.Get                (getByteString, getWord32le,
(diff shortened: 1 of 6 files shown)


m0dev76:
I guess the changes were to reflect the code in bitcoin core?

https://github.com/bitcoin/bitcoin/blob/master/src/arith_uint256.cpp#L206

haskoin-core/src/Network/Haskoin/Block/Types.hs
```diff
@@ -339,28 +335,51 @@ instance Serialize Headers where
...
-    c1 | s1 < 3    = posi `shiftL` (8*(3 - s1))
-       | otherwise = posi `shiftR` (8*(s1 - 3))
-    (s2,c2) | c1 .&. 0x00800000 /= 0  = (s1 + 1, c1 `shiftR` 8)
-            | otherwise               = (s1, c1)
-    c3 = fromIntegral $ c2 .|. (toInteger s2 `shiftL` 24)
-
+    i' = abs i
+    neg = i < 0
+    nSize' :: Int
+    nSize' = let f n = let n' = n `shiftR` 8
```
m0dev76: According to https://github.com/bitcoin/bitcoin/blob/master/src/arith_uint256.cpp#L228 `nSize == 0` is a valid possibility but this code will always give at least `1`. Maybe consider something like this:

```
nSize' = let f n = if n == 0 then 0 else 1 + f (n `shiftR` 8)
          in f i' 
```

haskoin-core/test/Network/Haskoin/Crypto/Hash/Units.hs
```diff
@@ -40,31 +36,22 @@ mapDRBGRsd vs = testCase "HMAC DRBG Vectors" $ mapM_ testDRBGRsd $ zip vs [0..]
...
-    let w1 = hmacDRBGNew (v !! 0) (v !! 1) (v !! 2)
+    let w1 = hmacDRBGNew (head v) (v !! 1) (v !! 2)
         w2 = hmacDRBGRsd w1 (v !! 3) (v !! 4)
         (w3,_) = hmacDRBGGen w2 128 (v !! 5)
         (_,r)  = hmacDRBGGen w3 128 (v !! 6)
     assertBool name $ fromJust r == (v !! 7)
     where v = map (fromJust . decodeHex) s
-          name = "    > HMAC DRBG Vector " ++ (show i)
-
-testCompact :: Assertion
```
m0dev76, line 60: Why remove these tests? Are they now invalid? If so maybe replace them with new valid ones?

m0dev76 (changes requested):

haskoin-core/src/Network/Haskoin/Block/Types.hs
jprupp: That was an excellent catch sir.

Image of a big fish

I added new test vectors that fail with the buggy code, as well as a fix for this issue.

haskoin-core/test/Network/Haskoin/Crypto/Hash/Units.hs
jprupp: Tests have been moved. They still exist.

m0dev76 (approved):

PR #340, opened 2018-01-07
Handle transaction fees in bytes

m0dev76:
* Transaction fees are now handled in bytes instead of kbytes.

* Transaction fee guessing was always assuming 2 outputs (one recipient
and one change). This number of outputs can now be passed as an argument.

### DIFF
--- a/haskoin-core/src/Network/Haskoin/Transaction/Builder.hs
+++ b/haskoin-core/src/Network/Haskoin/Transaction/Builder.hs
@@ -1,23 +1,5 @@
 {-# LANGUAGE OverloadedStrings #-}
-module Network.Haskoin.Transaction.Builder
-( Coin(..)
-, buildTx
-, buildAddrTx
-, SigInput(..)
-, signTx
-, signInput
-, mergeTxs
-, verifyStdTx
-, verifyStdInput
-, guessTxSize
-, chooseCoins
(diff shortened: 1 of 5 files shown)


haskoin-core/src/Network/Haskoin/Transaction/Builder.hs
```diff
@@ -160,17 +149,13 @@ greedyAddSink target fee continue =
...
+    byteFee * (fromIntegral $ guessTxSize nIn [] nOut 0)
 
-getMSFee :: Word64 -> (Int, Int) -> Int -> Word64
-getMSFee kbfee ms count =
-    kbfee*((len + 999) `div` 1000)
-  where
-    len = fromIntegral $ guessTxSize 0 (replicate count ms) 2 0
+guessMSTxFee :: Word64 -> (Int, Int) -> Int -> Int -> Word64
+guessMSTxFee byteFee ms nOut nIn =
+    byteFee * (fromIntegral $ guessTxSize 0 (replicate nIn ms) nOut 0)
```
jprupp: Remove parentheses:

```haskell
byteFee * fromIntegral (guessTxSize 0 (replicate nIn ms) nOut 0)
```

jprupp (approved):

PR #342, opened 2018-01-12
Remove redundant functions

jprupp:


haskoin-core/src/Network/Haskoin/Crypto/Base58.hs
```diff
@@ -110,10 +110,10 @@ instance Serialize Address where
               | otherwise = fail "Does not recognize address prefix"
     put (PubKeyAddress h) = do
         putWord8 addrPrefix
-        putByteString (hash160ToBS h)
+        putByteString (encode h)
```
m0dev76: Reuse the `Serialize` instances here for Hash160 (`put h`). Same for the `get` instance above.

m0dev76 (approved):

PR #343, opened 2018-01-15
Expose hash constructors

jprupp:
I need some of these constructors exposed as opposed to just accessors.


jprupp:
Encourages type-unsafe data.

PR #345, opened 2018-01-19
Change how scripts and addresses relate

jprupp:
Do not reuse the `Address` inside the lower-level `OutputScript` type, but use the embedded `Hash160`.

### DIFF
--- a/haskoin-core/src/Network/Haskoin/Script/Parser.hs
+++ b/haskoin-core/src/Network/Haskoin/Script/Parser.hs
@@ -140,16 +140,10 @@ addressToScriptBS = encode . addressToScript
 
 -- | Encode an output script as an address if it has such representation.
 scriptToAddress :: Script -> Maybe Address
-scriptToAddress s =
-    case decodeOutput s of
-        Left _ -> Nothing
-        Right output -> eitherToMaybe (outputAddress output)
+scriptToAddress = eitherToMaybe . (outputAddress <=< decodeOutput)
 
 scriptToAddressBS :: ByteString -> Maybe Address
-scriptToAddressBS s =
-    case decodeOutputBS s of
-        Left _ -> Nothing
-        Right output -> eitherToMaybe (outputAddress output)
+scriptToAddressBS = eitherToMaybe . (outputAddress <=< decodeOutputBS)
 
 -- | Computes a 'Script' from a 'ScriptOutput'. The 'Script' is a list of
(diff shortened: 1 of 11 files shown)


haskoin-core/src/Network/Haskoin/Test/Script.hs
```diff
@@ -222,7 +222,7 @@ arbitraryPKOutput =  PayPK . snd <$> arbitraryPubKey
 
 -- | Arbitrary ScriptOutput of type PayPKHash
 arbitraryPKHashOutput :: Gen ScriptOutput
-arbitraryPKHashOutput = PayPKHash <$> arbitraryPubKeyAddress
+arbitraryPKHashOutput = PayPKHash . getAddrHash <$> arbitraryPubKeyAddress
```
m0dev76: We can use `arbitraryHash160` here now

haskoin-core/src/Network/Haskoin/Transaction/Builder.hs
```diff
@@ -202,8 +202,8 @@ buildAddrTx xs ys =
     buildTx xs =<< mapM f ys
   where
     f (s, v) = case base58ToAddr s of
-        Just a@(PubKeyAddress _) -> return (PayPKHash a,v)
-        Just a@(ScriptAddress _) -> return (PayScriptHash a,v)
+        Just (PubKeyAddress h) -> return (PayPKHash h, v)
+        Just (ScriptAddress h) -> return (PayScriptHash h, v)
```
m0dev76: Here you could use your new `addressToScript` function

haskoin-core/src/Network/Haskoin/Script/Parser.hs
```diff
@@ -121,19 +126,40 @@ sortMulSig out = case out of
...
+-- | Encode address as output script in ByteString form.
+addressToScriptBS :: Address -> ByteString
+addressToScriptBS = encode . addressToScript
+
+-- | Encode an output script as an address if it has such representation.
+scriptToAddress :: Script -> Maybe Address
+scriptToAddress s =
+    case decodeOutput s of
+        Left _ -> Nothing
+        Right output -> eitherToMaybe (outputAddress output)
```
m0dev76: why not `scriptToAddress = eitherToMaybe . (outoutAddress <=< decodeOutput)`

m0dev76 (approved):

PR #348, opened 2018-02-07
nit: export isDataCarrier in Network.Haskoin.Script

hudon:
The check on whether the type of a `ScriptOutput` is a `DataCarrier` was missing from the exports in `Script.hs`


m0dev76:
Thanks!

PR #349, opened 2018-02-11
Adding testnet3 configuration

tgass:
Running hw using testnet doesn't work currently. I added testnet3 confiugration to config.yml. Is this actually the way to go? 

tilmann$ stack exec hw -- -t start
hw: DB config settings for testnet3 not found
CallStack (from HasCallStack):
  error, called at src/Network/Haskoin/Wallet/Server.hs:218:14 in haskoin-wallet-0.4.3-BbXtrCLDaLjCTNiaMs4Fql:Network.Haskoin.Wallet.Server

### DIFF
--- a/haskoin-wallet/config/config.yml
+++ b/haskoin-wallet/config/config.yml
@@ -20,6 +20,9 @@ database:
     testnet5:
         database: hw-wallet.sqlite3
         poolsize: 1
+    testnet3:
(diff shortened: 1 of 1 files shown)


jprupp:
Old wallet has been deprecated, but feel free to work on the [old-haskoin-wallet](https://github.com/haskoin/old-haskoin-wallet) package if you want to keep this going.

PR #354, opened 2018-07-27
Upgrading haskoin-core to GHC 8.4.3

wraithm:
I also have a patch for `haskoin-wallet` and `haskoin-node`, but `esqueleto` hasn't fully upgraded to GHC 8.4.3 yet, so I'm gonna wait on that.

### DIFF
--- a/haskoin-core/haskoin-core.cabal
+++ b/haskoin-core/haskoin-core.cabal
@@ -72,24 +72,23 @@ library
                          Network.Haskoin.Test.Transaction
                          Network.Haskoin.Test.Block
                          Paths_haskoin_core
-    build-depends:       aeson >= 1.2 && < 1.3
+    build-depends:       aeson >= 1.2 && < 1.4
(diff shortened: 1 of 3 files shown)


wraithm:
@xenog @runeksvendsen @plaprade, any thoughts on this?

runeksvendsen:
Does this break earlier lts-versions,
or is it backwards compatible?

jprupp:
Both @plaprade and I have been working on non-backwards-compatible changes to Haskoin.

The SPV wallet is being rewritten into a server (Haskoin Store), which is a Bitcoin protocol client that stores the entire indexed blockchain, and a simple command-line wallet that consumes data from that store over HTTP/JSON. Notably Bitcoin Cash is being supported, and a bunch of smaller changes are in the pipeline.

The new Haskoin does not use Esqueleto and is compatible with the latest LTS Haskell version. It will be merged soon®.

PR #356, opened 2018-08-23
Swap s and r in Tresor RFC 6979 test vectors

ssadler:
Hi again,

The previous fix to swap s and r which was merged and subsequently reverted broke these tests; as far as I can tell these vectors don't exist anywhere online except this test repo, as the Trezor Python ecdsa repository has since disappeared. 

To clarify, the issue I had is that S and R are swapped inside CompactRecSig.

Thanks again for this library I've had alot of use out of it.


ssadler:
I am wrong again! The vectors are correct according to secp256k1..

Issue #358, opened 2018-09-04
Using Text instead of ByteString for human readable strings


wraithm:
I love the work that's going into haskoin-core 1.0.0! This is amazing. My only complaint is that functions like `txHashToHex :: TxHash -> ByteString` and `addrToString :: Address -> Maybe ByteString` use implicitly utf8-encoded `ByteString`s instead of `Text`. I end up using wrappers around these functions in my code. It would be nice if those functions just did `Text` directly because they're explicitly utf encoded human-readable strings, which is what `Text` is for.

What does everybody think about this?

wraithm:
@xenog @plaprade

jprupp:
The reason I used `ByteString` for addresses and hex-encoded strings is that they comprise only characters that fit in an eight-bit word (low Unicode). It is thus wasteful to use `Text` or `[Char]`, as those take up at least twice as much memory (16 or 32 bits) to represent every single character, plus the required overhead of convertion when needed.

I was trying to optimise the code as much as I could back when I chose `ByteString` for these data types, but I am now convinced that the string representations and conversion for these types are not performance bottlenecks that matter, and I would gladly switch `Text` if that is convenient.

wraithm:
I'd be happy to make a PR for it.

wraithm:
@xenog I just tried to clone `haskoin-core` to build, and I got this error on building `secp256k1`:

```secp256k1-1.1.2: configure (lib)
Configuring secp256k1-1.1.2...
Cabal-simple_mPHDZzAJ_2.2.0.1_ghc-8.4.3: Missing dependency on a foreign
library:
* Missing (or bad) C library: secp256k1
This problem can usually be solved by installing the system package that
provides this library (you may need the "-dev" version). If the library is
already installed but in a non-standard location then you can use the flags
--extra-include-dirs= and --extra-lib-dirs= to specify where it is.If the
library file does exist, it may contain errors that are caught by the C
compiler at the preprocessing stage. In this case you can re-run configure
with the verbosity flag -v3 to see the error messages.
```

Do I need to have the C library `secp256k1` installed globally to use `haskoin-core`? What do you do for your dev setup?

sha49:
I faced the same issues today. 

Yes, secp256k1 C is a dependency of haskoin core. So, you can clone the repo https://github.com/bitcoin-core/secp256k1 and follow the build steps or you can install `nix` on your machine and then run `stack --nix install` and then the dependencies will be set locally by nix.

[...]

PR #360, opened 2018-09-06
Converting human-readable strings to Text instead of ByteString

wraithm:
This is in response to issue #358.

### DIFF
--- a/haskoin-core.cabal
+++ b/haskoin-core.cabal
@@ -2,7 +2,7 @@
 --
 -- see: https://github.com/sol/hpack
 --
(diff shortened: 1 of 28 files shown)


wraithm:
All of the tests pass! Seems to work well.

wraithm:
@xenog, I hope you don't mind. I made some minor cleanup changes. I just removed some redundant imports, used `zipWithM_` instead of `sequence_ . zipWith ...`, and things like that.

wraithm:
Actually, don't merge this yet. I just found that there are a bunch of redundant calls to `cs` that we can get rid of.

wraithm:
Okay, I've removed as many redundant cs calls as I could find.

PR #364, opened 2018-11-02
PSBT data types and serialization

wraithm:
This creates the data types and serializations for Partially Signed Bitcoin Transactions, [BIP-174](https://github.com/bitcoin/bips/blob/master/bip-0174.mediawiki).

The next step is to write signature merging code.


wraithm:
@xenog @plaprade Any thoughts?

jprupp:
I think it is very interesting. I will do a full review and if it looks good merge it. Thank you for this. It is very good to have.

PR #368, opened 2018-12-05
Cleanups and removing warnings

wraithm:
Just some minor stuff that removes some warnings that I accidentally introduced.


wraithm:
@xenog, any thoughts?

PR #374, opened 2019-04-23
Expose transaction signature

GambolingPangolin:
This PR adds a function which computes a `TxSignature` over a transaction in case detached signatures are needed for a transaction construction workflow.


src/Network/Haskoin/Transaction/Builder.hs
```diff
@@ -45,19 +46,14 @@ import           Control.Arrow                      (first)
 import           Control.Monad                      (foldM, mzero, unless, when)
 import           Control.Monad.Identity             (runIdentity)
 import           Crypto.Secp256k1
-import           Data.Aeson                         (FromJSON, ToJSON,
-                                                     Value (Object), object,
-                                                     parseJSON, toJSON, (.:),
+import           Data.Aeson                         (FromJSON, ToJSON, Value (Object), object, parseJSON, toJSON, (.:),
```
wraithm: 80 character line limit

PR #378, opened 2019-10-26
lowercase input before bech32 encoding

### LINKED ISSUE #375: `bech32Encode` generates invalid output if the HRP contains upper case characters.
## Overview

When [`bech32Encode`](https://github.com/haskoin/haskoin-core/blob/d14b988dcfd96e20ec260ff8fd2f45f288c8587f/src/Network/Haskoin/Address/Bech32.hs#L131) is called with a human-readable part that contains one or more **upper-case characters**, it produces in an **invalid** Bech32 string that cannot be decoded with the `bech32Decode` function.

## Analysis

As part of the encoding process, the human-readable part is converted to lower case:

https://github.com/haskoin/haskoin-core/blob/d14b988dcfd96e20ec260ff8fd2f45f288c8587f/src/Network/Haskoin/Address/Bech32.hs#L136

However, the checksum is calculated **before** the conversion to lower case takes place:

https://github.com/haskoin/haskoin-core/blob/d14b988dcfd96e20ec260ff8fd2f45f288c8587f/src/Network/Haskoin/Address/Bech32.hs#L134

This contradicts the [Bech32 specification](https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#bech32), which states:

> "The lowercase form is used when determining a character's value for checksum purposes."

Therefore, if the original human-readable part contains one or more upper case characters:

1. the generated checksum will be inconsistent with the human-readable prefix of the output string
2. the output string will fail to decode.

## Example

Consider the following two calls to `bech32Encode`, differing only in the case of the human-readable part:
```hs
> bech32Encode "test" []
> bech32Encode "TEST" []
```
### Expected Behaviour

Both calls to `bech32Encode` should result in the **same** output string:
```hs
> bech32Encode "test" []
Just "test12hrzfj"
> bech32Encode "TEST" []
Just "test12hrzfj"
> bech32Encode "test" [] == bech32Encode "TEST" [] 
True
```

### Actual Behaviour

The above calls to `bech32Encode` **actually** result in **different** output strings:
```hs
> bech32Encode "test" []
Just "test12hrzfj"
> bech32Encode "TEST" []
Just "test13jgcyw"
> bech32Encode "test" [] == bech32Encode "TEST" [] 
False
```
Attempting to decode the string produced by `bech32Encode "TEST" []` results in `Nothing`:
```hs
> bech32Decode "test13jgcyw"
Nothing
```

ai-pavel:
fixes https://github.com/haskoin/haskoin-core/issues/375

Quick shortcut for testing:

    $ stack test --test-arguments "--match=bech32"

### DIFF
--- a/src/Network/Haskoin/Address/Bech32.hs
+++ b/src/Network/Haskoin/Address/Bech32.hs
@@ -131,7 +131,7 @@ maxBech32Length = 90
 bech32Encode :: HRP -> [Word5] -> Maybe Bech32
 bech32Encode hrp dat = do
     guard $ checkHRP hrp
-    let dat' = dat ++ bech32CreateChecksum hrp dat
+    let dat' = dat ++ bech32CreateChecksum (T.toLower hrp) dat
         rest = map (charset !) dat'
         result = T.concat [T.toLower hrp, T.pack "1", T.pack rest]
     guard $ T.length result <= maxBech32Length
--- a/test/Network/Haskoin/Address/Bech32Spec.hs
+++ b/test/Network/Haskoin/Address/Bech32Spec.hs
@@ -28,6 +28,8 @@ spec = do
         it "should be a valid address" $ forM_ validAddresses testValidAddress
         it "should be an invalid address" $
             forM_ invalidAddresses testInvalidAddress
+        it "should be same for the input data in different case" $
+            all (== Just "test12hrzfj") (map bech32EncodeDefault hrpCaseVariants)
     describe "more encoding/decoding cases" $ do
         it "length > 90" $
             assert $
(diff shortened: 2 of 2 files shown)


runeksvendsen:
In my opinion, if uppercase text is not defined for a human-readable representation of a bech32 address, then the data type used to represent this should not be able to contain uppercase characters. So, instead of using a type synonym `type HRP = Text`, we should use a `newtype` with a smart contractor that performs any necessary conversions (such as lower-casing) upon construction. This avoids requiring that every function, that takes a `HRP` as an argument, do this manually.

jprupp:
Technically the HRP is defined as case-insensitive, but there is nothing to prevent it from being case-_preserving_, so it seems correct to me to use `Text` instead of a custom type.

PR #381, opened 2020-01-14
adds cabal project files

GambolingPangolin:
I added the configuration required for this to also be a cabal project. This includes a "freeze" file which pins package versions to those in lts-14.19 (currently set by `stack.yaml`).


wraithm:
imho, we should just add the freeze file to the gitignore and then maybe a script or something to download it from the stack.yaml? Still commit the project file.

What do you think @xenog?

GambolingPangolin:
The reason for including a freeze file is so that a cabal build uses the same dependencies as a stack build.  This should make it more likely that build problems will come up for stack builds iff they come up for cabal builds.  It might not be necessary to include a `cabal.project` at all since there is no configuration in it.

runeksvendsen:
@GambolingPangolin It appears to me that the freeze file can be deduced from the stack.yaml file. Is this correct? And, if so, why include it in the repository?

wraithm:
It's just more stuff to update every time we bump versions. Headache. Just write a script to download the freeze file based off of the stack.yaml resolver.

GambolingPangolin:
@runeksvendsen Even better, Stackage makes a freeze file for each snapshot available at `https://stackage.org/lts-XX/cabal.config`.  In our case we need to remove the entry for `haskoin-core`.

GambolingPangolin:
I think adding instructions (to the README?) on how to get a freeze file for the current snapshot would be better than including a script.

jprupp:
I see there is discussion here. I don't want to include more files to maintain when there is a version change. The build toolset change from `cabal-install` to `stack` makes sense, and I wouldn't like to continue supporting legacy `cabal-install`, not to be confused with the Cabal library, which is still widely used. I won't merge this.

wraithm:
Agreed.

PR #384, opened 2020-05-16
Add better JSON instances and cleanup some code

m0dev76:
The only functional changes in this commit are changes to the JSON instances to provide meaningful defaults. There's also a bit of aesthetic code cleanup and "space m f".

### DIFF
--- a/src/Haskoin/Block/Common.hs
+++ b/src/Haskoin/Block/Common.hs
@@ -35,9 +35,10 @@ import           Control.Monad              (forM_, liftM2, mzero, replicateM)
 import           Data.Aeson                 (FromJSON (..), ToJSON (..),
                                              Value (..), object, toJSON,
                                              withObject, withText, (.:), (.=))
-import           Data.Aeson.Encoding        (pairs, text)
+import           Data.Aeson.Encoding        (pairs, unsafeToEncoding)
 import           Data.Bits                  (shiftL, shiftR, (.&.), (.|.))
 import qualified Data.ByteString            as B
+import           Data.ByteString.Builder    (char7)
 import           Data.Hashable              (Hashable)
 import           Data.Maybe                 (fromMaybe)
 import           Data.Serialize             (Serialize, decode, encode, get,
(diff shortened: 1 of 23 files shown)


src/Haskoin/Transaction/Common.hs
```diff
@@ -75,9 +74,7 @@ instance FromJSON TxHash where
 
 instance ToJSON TxHash where
     toJSON = A.String . txHashToHex
-    toEncoding (TxHash h) =
-        unsafeToEncoding $
-        char7 '"' <> hexBuilder (B.reverse (S.encode h)) <> char7 '"'
+    toEncoding h = text $ txHashToHex h
```
jprupp: This is pretty but not efficient.

src/Haskoin/Block/Common.hs
```diff
@@ -128,14 +107,12 @@ instance IsString BlockHash where
...
-    parseJSON = withText "block hash" $
+    parseJSON = withText "BlockHash" $
         maybe mzero return . hexToBlockHash
 
 instance ToJSON BlockHash where
     toJSON = String . blockHashToHex
-    toEncoding s =
-        unsafeToEncoding $
-        char7 '"' <> hexBuilder (B.reverse (encode s)) <> char7 '"'
+    toEncoding = text . blockHashToHex
```
jprupp: Pretty but not efficient.

src/Haskoin/Keys/Common.hs
```diff
@@ -66,7 +65,7 @@ instance IsString PubKeyI where
 
 instance ToJSON PubKeyI where
     toJSON = String . encodeHex . encode
-    toEncoding s = unsafeToEncoding $ char7 '"' <> hexBuilder (encode s) <> char7 '"'
+    toEncoding = text . encodeHex . encode
```
jprupp: Again, pretty but not efficient.

jprupp (changes requested):
Revert those changes to the encodings for `TxHash` and `BlockHash` because they are fast,and I need to encode large numbers of `TxHash` and `BlockHash` structures from Haskoin Store.

PR #387, opened 2020-06-04
Expose the Arbitrary test instances under Haskoin.Util.Arbitrary

m0dev76:
Title says it all. It's simply to avoid duplicating work in downstream code.


jprupp:
Do we need `Util` there in the path?

PR #392, opened 2020-07-03
renames OP_CLTV and OP_CSV opcodes

GambolingPangolin:
CLTV and CSV have been part of bitcoin consensus since 2016.  I renamed the relevant opcodes to reflect this reality.  This is a breaking change, but it seems very unlikely that anyone was actually using OP_NOP2 and OP_NOP3 directly.  So I feel that changing the constructor names is an acceptable approach rather than something more conservative.


jprupp:
Thank you, good catch.

jprupp (approved):

PR #395, opened 2020-10-31
Adds support for base16-bytestring >=1

GambolingPangolin:
`base16-bytestring` recently changed the signature of the decoding function for version 1.0.  This change makes `haskoin-core` compatible with versions up to and including 1.0.


wraithm (approved):
LGTM

GambolingPangolin:
I think it would also make sense to make revisions on Hackage with `base16-bytestring >= 0.1.1.6 && < 1.0.0` for `haskoin-core` versions prior to this one (0.17.1) going back to 0.13.

PR #398, opened 2021-10-03
Change representation for fingerprints

GambolingPangolin:
This changes how we represent fingerprints to `newtype Fingerprint = Fingerprint { unFingerprint :: Word32 }`.  This has a few benefits.  First, `Fingerprint` has `Serialize` etc. instances and downstream structures use these instances, so there is no risk of the encoding drifting from one place to another.  (The binary encodings for fingerprints in the extended keys section disagreed with that from the psbt section.)   Second, I added IsString/Read/Show instances that treat `Fingerprint` as having an 8-digit hex string representation instead of a decimal number.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## 0.21.0
+### Fixed
+- Make fingerprint serialization agree everywhere
+
 ## 0.20.5
 ### Added
 - Support Bech32m address format for Taproot.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## 0.21.0
-### Fixed
-- Make fingerprint serialization agree everywhere
+### Changed
+- Use a newtype for Fingerprint, which uses an 8 digit hex string for IsString,
+  Read, & Show.  This fixes inconsistent (de)serialization across the package.
 
 ## 0.20.5
 ### Added
--- a/haskoin-core.cabal
+++ b/haskoin-core.cabal
@@ -1,10 +1,10 @@
 cabal-version: 1.12
 
--- This file has been generated from package.yaml by hpack version 0.34.4.
+-- This file has been generated from package.yaml by hpack version 0.34.5.
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 83e95434961483bbb306908fbc61269ebec705003a8994ca0a8a04d1e3f7de50
+-- hash: 4b340a6dfe197ae20194205ecebc4358218dfb13d1451ff7606ed6290a4ba10e
 
 name:           haskoin-core
 version:        0.20.5
@@ -86,7 +86,7 @@ library
       Haskoin.Util.Arbitrary.Transaction
       Haskoin.Util.Arbitrary.Util
   other-modules:
-      Paths_haskoin_core
+      Haskoin.Keys.Extended.Internal
   hs-source-dirs:
       src
   build-depends:
--- a/src/Haskoin/Keys/Extended/Internal.hs
+++ b/src/Haskoin/Keys/Extended/Internal.hs
@@ -1,0 +1,60 @@
+{-# LANGUAGE DeriveGeneric              #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE TypeApplications           #-}
+
+module Haskoin.Keys.Extended.Internal (
+    Fingerprint (..)
+) where
+
+import           Control.DeepSeq   (NFData)
+import           Data.Binary       (Binary (..))
+import           Data.Bytes.Get    (getWord32be)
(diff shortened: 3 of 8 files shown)


src/Haskoin/Transaction/Partial.hs
```diff
@@ -679,15 +679,15 @@ instance Serialize PSBTHDPath where
...
+                ((,) <$> S.getWord32be <*> getKeyIndexList numIndices)
       where
         getKeyIndexList n = replicateM n S.getWord32le
 
     put (PSBTHDPath (fp, kis)) = do
         putVarInt (B.length bs)
         S.putByteString bs
       where
-        bs = S.runPut $ S.putWord32le fp >> mapM_ S.putWord32le kis
+        bs = S.runPut $ S.putWord32be fp >> mapM_ S.putWord32le kis
```
wraithm: What about the other word32le there?

Is there mixed byteorder?

Also, I'm surprised the test suite doesn't catch this!!

Would you mind checking if the psbt test suite is missing vectors?

GambolingPangolin: We might also need to change that.  `DerivPathI` uses big-endian byte order for key indices.  I'll add test vectors as needed and make sure that what we do conforms with the BIP 174 spec.  The BIP has a lot of test vectors.

GambolingPangolin: The situation is messy:
- BIP 32 specifies big-endian for key indices
- BIP 174 specifies little-endian for HD path components

Also, I added a test case covering HD paths.

GambolingPangolin:
Another possibility is to put `Fingerprint` behind an abstract `newtype`, using `Word32` as an opaque implementation detail.  Then if downstream (de)serializers used its serialization instances there would be no danger of divergence.  Also, we could make the show instance be the hex representation.

test/Haskoin/Transaction/PartialSpec.hs
```diff
@@ -137,6 +203,7 @@ vec6Test = do
         (Key 0x0f (fromJust $ decodeHex "010203040506070809"))
         (fromJust $ decodeHex "0102030405060708090a0b0c0d0e0f")
 
+
```
wraithm: nit: unnecessary whitespace

test/Haskoin/Transaction/PartialSpec.hs
```diff
@@ -121,6 +119,74 @@ vec5Test = do
...
+                   ]
+        , outputs = [ Output { outputRedeemScript = Nothing
+                             , outputWitnessScript = Nothing
+                             , outputHDKeypaths = mempty
+                             , outputUnknown = mempty
+                             }
+                    ]
+        }
+    hardIndex = (+ 2^31)
+
```
wraithm, line 190: nit: unnecessary whitespace

GambolingPangolin:
I went ahead and changed `Fingerprint` to a `newtype`.

haskoin-core.cabal
```diff
@@ -86,7 +86,7 @@ library
       Haskoin.Util.Arbitrary.Transaction
       Haskoin.Util.Arbitrary.Util
   other-modules:
-      Paths_haskoin_core
```
wraithm, line 89: Why remove this?

wraithm: Oh, it's automatic! hmm

src/Haskoin/Keys/Extended/Internal.hs
```diff
@@ -0,0 +1,61 @@
...
+instance IsString Fingerprint where
+    fromString = fromRight decodeError
+        . S.decode
+        . fromMaybe hexError
+        . decodeHex
+        . Text.pack
+     where
+       decodeError  = error "Fingerprint literal: Unable to decode"
+       hexError = error "Fingerprint literal: Invalid hex"
+
```
wraithm, line 49: nit: 2 newlines

haskoin-core.cabal
GambolingPangolin: This must have been a side effect of _adding_ an `other-modules` field to `package.yaml`.  Do we use the paths module for anything?  The project builds without it.  Maybe we should leave it off.

wraithm: Yeah, let's maybe leave it off. The paths modules cause all kinds of problems with test rebuilding and stuff... Would you mind actually explicitly ignoring it (like we do in our repo)?

GambolingPangolin: Sure!

CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
+## 0.21.0
+### Changed
+- Use a newtype for Fingerprint, which uses an 8 digit hex string for IsString,
+  Read, & Show.  This fixes inconsistent (de)serialization across the package.
```
wraithm, line 10: It might be good to mention the Paths_haskoin_core thing.

@jprupp, what do you think about this? I don't imagine that people are depending on this module, but you can tell me otherwise.

wraithm: We can also take the conservative route and still expose it.

GambolingPangolin: `Paths_haskoin_core` was not part of the public API.  It was under `other-modules` and now it's unlisted altogether.

wraithm: Ah, okay! Nvm then. Resolving.

wraithm (approved):

wraithm:
@jprupp, Do you want to take a look at this? I confirmed that this is now ready to go.

jprupp:
Thanks @wraithm. Looks good to me.

PR #402, opened 2021-10-26
Adds PSBT signing capability

GambolingPangolin:


src/Haskoin/Transaction/Partial.hs
```diff
@@ -185,6 +206,152 @@ mergeOutput a b = Output
...
+    }
+
+instance Semigroup PsbtSigner where
+    PsbtSigner signer1 <> PsbtSigner signer2 =
+        PsbtSigner $ \pubKey origin ->
+            signer1 pubKey origin <|> signer2 pubKey origin
+
+instance Monoid PsbtSigner where
+    mempty = PsbtSigner $ \_ _ -> Nothing
+
```
wraithm, line 228: nit: whitespace

@jprupp, what do you think about adding a formatter that solves this? eg. ormolu or fourmolu?

src/Haskoin/Transaction/Partial.hs
```diff
@@ -185,6 +206,152 @@ mergeOutput a b = Output
...
+    { unPsbtSigner ::
+        PubKeyI ->
+        Maybe (Fingerprint, DerivPath) ->
+        Maybe SecKey
+    }
+
+instance Semigroup PsbtSigner where
+    PsbtSigner signer1 <> PsbtSigner signer2 =
+        PsbtSigner $ \pubKey origin ->
+            signer1 pubKey origin <|> signer2 pubKey origin
```
wraithm, line 224: I don't totally understand what's going on here. You're just selecting *any* signer?

GambolingPangolin: Each signer corresponds to a set secret keys that you look up using a `PubKeyI` and optional fingerprint-path pair.  The Monoid instance is union in this view.

src/Haskoin/Transaction/Partial.hs
GambolingPangolin: I addressed this but am leaving it open in case @jprupp wants to weigh in.  Also see https://github.com/haskoin/haskoin-core/issues/399

jprupp: I used to be a bit more strict with formatting, but now that I'm getting old, I don't care as much. As long as the code is clean and understandable, I don't care how you format it. If you'd like to implement a formatting guideline for this package, and can document how to get it working on Spacemacs, I'm fine with adopting it.

jprupp (approved):

wraithm (approved):

